PR #51, opened 2018-06-20
Make running the setup script easier, in response to actual user experience.

pboothe:
Only source the bash.inc file if gcloud is not on the path.

Update the docs to make sure people know about needing to run things from the right directory.

This change is [Reviewable](https://reviewable.io/reviews/m-lab/k8s-support/51)

### DIFF
--- a/cloud/README.md
+++ b/cloud/README.md
@@ -2,15 +2,18 @@ All setup and startup scripts for cloud and cloud networking.
 
 The scripts need to be run from this directory.
 
-To set up a cloud master in sandbox, run ```bash
+To set up a cloud master in sandbox, run
+```bash
 ./setup_cloud_k8s_master.sh
 ```
 
-To set up a cloud master in staging, run ```bash
+To set up a cloud master in staging, run
+```bash
 GOOGLE_CLOUD_PROJECT=mlab-staging ./setup_cloud_k8s_master.sh
 ```
 
-To set up a cloud master in staging, run ```bash
+To set up a cloud master in staging, run
+```bash
 GOOGLE_CLOUD_PROJECT=mlab-oti ./setup_cloud_k8s_master.sh
 ```
 
--- a/cloud/README.md
+++ b/cloud/README.md
@@ -4,34 +4,32 @@ The scripts need to be run from this directory.
 
 To set up a cloud master in sandbox, run
 ```bash
-./setup_cloud_k8s_master.sh
+./setup_cloud_k8s_master.sh mlab-sandbox
 ```
 
 To set up a cloud master in staging, run
 ```bash
-GOOGLE_CLOUD_PROJECT=mlab-staging ./setup_cloud_k8s_master.sh
+./setup_cloud_k8s_master.sh mlab-staging
 ```
 
 To set up a cloud master in production, run
 ```bash
-GOOGLE_CLOUD_PROJECT=mlab-oti ./setup_cloud_k8s_master.sh
+./setup_cloud_k8s_master.sh mlab-oti
 ```
 
 # Master node setup
 
 We use `kubeadm` to set everything up.  It's alpha, but it works pretty well.
-The `setup_cloud_k8s_master.sh` script, when run with no arguments, sets up
-`k8s-platform-master` in the `mlab-sandbox` project.
 
 # Network
 
 Our network is non-standard, because kubernetes does not expect to expose
 services running on pods directly to the outside world.  So, for the pods that
-run services we want to expose (running on platform nodes), we actually give
-them two IP addresses using [multus](https://github.com/intel/multus-cni) to
-speciy two interfaces instead of just one.  One internal one, handed out with
-[flannel](https://github.com/coreos/flannel) in the standard way flannel does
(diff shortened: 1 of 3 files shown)


nkinkade:
Comments added.

---

Review status: 0 of 1 LGTMs obtained

---

*[cloud/README.md, line 7 at r1](https://reviewable.io/reviews/m-lab/k8s-support/51#-LFT4ZRC3Qzs9LjaqRER:-LFT4ZRC3Qzs9LjaqRES:b-2jds55) ([raw file](https://github.com/m-lab/k8s-support/blob/19fd9b8ff06a9a9b0e7a3b193bff3d55fd21d3ec/cloud/README.md#L7)):*
> ```Markdown
> To set up a cloud master in sandbox, run
> ```bash
> ./setup_cloud_k8s_master.sh
> ```

Rather than having a default project, would it be better to throw an error noting that a project must be specified? Because the script blows away any existing instance of the VM in the given project, it seems better for this to be explicit rather than implicit. 

With this in mind, might it be more explicit to pass project as an argument, rather than accessing it through the environment? Most of our other support scripts take project as an argument. And, to me, this form is more explicit and straight forward:

`$ ./setup_cloud_k8s_master.sh mlab-oti`

---

*[cloud/setup_cloud_k8s_master.sh, line 21 at r1](https://reviewable.io/reviews/m-lab/k8s-support/51#-LFT5wxW2U4q3Sku0JuD:-LFT5wxW2U4q3Sku0JuE:b-o575ya) ([raw file](https://github.com/m-lab/k8s-support/blob/19fd9b8ff06a9a9b0e7a3b193bff3d55fd21d3ec/cloud/setup_cloud_k8s_master.sh#L21)):*
> ```Shell
> IP_NAME=${K8S_GCE_MASTER_IP:-k8s-platform-master-ip}
> 
> # Ensure gcloud is in the PATH.
> ```

This doesn't actually ensure that `gcloud` is in the user's `PATH`. For many users, including myself, `"${HOME}/google-cloud-sdk/path.bash.inc"` does not exist. Am I recalling correctly from our VC yesterday that your plan was to only run this `source` command when in a Travis-CI build environment? Maybe something like this instead:

```
if [[ "$TRAVIS" == "true" ]]; then
    source "${HOME}/google-cloud-sdk/path.bash.inc"
elif [[ ! $(which gcloud) ]]; then
    echo "The google-cloud-sdk must be installed and gcloud in your path."
    exit 1
fi
```

---


*Comments from [Reviewable](https://reviewable.io/reviews/m-lab/k8s-support/51#-:-LFT7RMd62qfjC2jwRu3:byymmtx)*

pboothe:
PTAL

---

Review status: 0 of 1 LGTMs obtained

---

*[cloud/README.md, line 7 at r1](https://reviewable.io/reviews/m-lab/k8s-support/51#-LFT4ZRC3Qzs9LjaqRER:-LFTPY0Y2iUufAjwmiXR:b-896fix) ([raw file](https://github.com/m-lab/k8s-support/blob/19fd9b8ff06a9a9b0e7a3b193bff3d55fd21d3ec/cloud/README.md#L7)):*

Rather than having a default project, would it be better to throw an error noting that a project must be specified? Because the script blows away any existing instance of the VM in the given project, it seems better for this to be explicit rather than implicit. 

With this in mind, might it be more explicit to pass project as an argument, rather than accessing it through the environment? Most of our other support scripts take project as an argument. And, to me, this form is more explicit and straight forward:

`$ ./setup_cloud_k8s_master.sh mlab-oti`
</blockquote></details>

Done.

---

*[cloud/setup_cloud_k8s_master.sh, line 21 at r1](https://reviewable.io/reviews/m-lab/k8s-support/51#-LFT5wxW2U4q3Sku0JuD:-LFTPYrh0SWIit7G8tYH:b3a0sdr) ([raw file](https://github.com/m-lab/k8s-support/blob/19fd9b8ff06a9a9b0e7a3b193bff3d55fd21d3ec/cloud/setup_cloud_k8s_master.sh#L21)):*

This doesn't actually ensure that `gcloud` is in the user's `PATH`. For many users, including myself, `"${HOME}/google-cloud-sdk/path.bash.inc"` does not exist. Am I recalling correctly from our VC yesterday that your plan was to only run this `source` command when in a Travis-CI build environment? Maybe something like this instead:

```
if [[ "$TRAVIS" == "true" ]]; then
    source "${HOME}/google-cloud-sdk/path.bash.inc"
elif [[ ! $(which gcloud) ]]; then
    echo "The google-cloud-sdk must be installed and gcloud in your path."
    exit 1
fi
```
</blockquote></details>

Fixed in a different way.

---


*Comments from [Reviewable](https://reviewable.io/reviews/m-lab/k8s-support/51#-:-LFTSOt_-fIV5HoS3PqK:bb71pro)*

nkinkade:
:lgtm:

---

Reviewed 3 of 3 files at r2.
Review status: :shipit: complete! 1 of 1 LGTMs obtained

---



*Comments from [Reviewable](https://reviewable.io/reviews/m-lab/k8s-support/51#-:-LFTctaoDA_ghPRrz0IH:bnfp4nl)*

PR #65, opened 2018-07-26
Restructure directories + new daemonsets and configs.

nkinkade:
This PR changes the directory structure of this repository significantly. It now more closely mirrors the tried-and-true structure of the m-lab/prometheus-support repository.

It also adds:

* A prometheus 2.x deployment and config, which is set to auto-discover targets.
* Adds a new daemonset for node_exporter.
* Adds two new demo node_exporter `textfile` collector files, which need to be added as configmaps, and then mounted into node_exporter's filesystem. These are temporary, simply for the purposes of the 10-year demo.
* Changes `apiversion` from `extensions/v1beta1` to `apps/v1` in various places, as this seems more current.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/k8s-support/65)


pboothe:
I would strongly prefer a subdirectory under k8s/ or k8s/daemonsets that splits the configs into "experiment" configs (which experiment developers might want to understand) and "internal" or "core_service" configs which are for us.

PR #67, opened 2018-07-26
Reuse kubeadm certs

stephen-soltesz:
This change adds support for reusing kubeadm certs stored in private GCS buckets associated with the current project.

GCS buckets storing the certs are mounted readonly at the default location /etc/kubernetes/pki. When kubeadm init runs, it automatically finds and reuses these certs.

```
[certificates] Using the existing ca certificate and key.
[certificates] Using the existing apiserver certificate and key.
[certificates] Using the existing apiserver-kubelet-client certificate and key.
...
[snip]
...
```

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/k8s-support/67)

### DIFF
--- a/cloud/setup_cloud_k8s_master.sh
+++ b/cloud/setup_cloud_k8s_master.sh
@@ -112,12 +112,35 @@ gcloud compute ssh "${GCE_ARGS[@]}" "${GCE_NAME}" <<-\EOF
   systemctl restart kubelet
 EOF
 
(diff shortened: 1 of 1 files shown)


stephen-soltesz:
@nkinkade fyi -- I've already setup gs://k8s-platform-master-mlab-staging an copied the staging credentials there. As far as I'm aware, the only requirement to deploy this change for staging would be to re-run the `setup_cloud_k8s_master.sh` script for mlab-staging.

PR #78, opened 2018-10-22
Fix the configs to be compatible with Multus 3.0

pboothe:
Multus (bless their hearts) completely changed their configs around.

The new configs look like they were designed by committee, which means that I suspect that the format will be stable for a long time.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/k8s-support/78)


pboothe:
Webhooks are apparently a little bit broken in GitHub right now.

pboothe:
PTAL

PR #81, opened 2018-10-24
Final updates to k8s master cluster script.

nkinkade:
This PR:
* Renames to the script setup_k8s_cloud_master.sh to bootstrap_k8s_master_cluster, which hopefully is more indicative of what it actually does.
* Adds a warning in the script that informs the user that the script is destructive and irreversible, asking if they are sure they want to continue.
* Removes all ePoxy bits, which will eventually be handled by the ePoxy code.
* Renames cloud.conf to cloud-provider.conf to be more clear about what the configuration file will (eventually) do.
* Cleans up the etcdctl env variables for root, and doesn't bother setting them up for a regular user, since that user doesn't have read access to various necessary key files.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/k8s-support/81)


pboothe:
LGTM, hooray!

PR #88, opened 2018-12-10
Increase sleep before restarting kubelet and between stop and start.

evfirerob:
This change is [Reviewable](https://reviewable.io/reviews/m-lab/k8s-support/88)


pboothe:
LGTM

PR #97, opened 2019-01-08
Don't scrape port 3010 on ndt-cloud nodes + adds toleration to ndt-cloud DaemonSet

nkinkade:
This change is [Reviewable](https://reviewable.io/reviews/m-lab/k8s-support/97)

### DIFF
--- a/config/prometheus/prometheus.yml
+++ b/config/prometheus/prometheus.yml
@@ -53,6 +53,7 @@ scrape_configs:
         target_label: __metrics_path__
         replacement: /api/v1/nodes/${1}/proxy/metrics
 
+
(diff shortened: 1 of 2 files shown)


nkinkade:
Prometheus should be scraping port 3010 of the ndt-cloud pods, which is the port where tests actually happen. It should only scrape port 9090.

Additionally, this PR adds a (hopefully) temporary toleration to the ndt-cloud pods which allows it to be schedule on pods with the `NetworkUnavailable=True` condition set (with corresponding taint). This is some issue with the router_controller complaining about not being able to create a route be "instance not found".

PR #147, opened 2019-03-22
Adds a new script for upgrading k8s (and etcd) on a master cluster.

nkinkade:
The primary change in this PR is a new script which can be used to upgrade all kubernetes components on an entire master cluster.

Additionally this PR includes:
* Bumping the configured k8s version to 1.13.4, and crictl to 0.7.5
* Improving the script to add a master node to an existing cluster. It is no longer necessary to pass in a "bootstrap" master zone, and instead the script just picks the first master zone in the list of zones for the GCP project that isn't the one being added.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/k8s-support/147)


nkinkade:
Reassigned to @stephen-soltesz because @pboothe won't have time to look over this fully today.

PR #166, opened 2019-04-02
Sets ipvlan mode to l3s, and adds a new flannel config for experiments.

nkinkade:
The default ipvlan mode is "l2". In this mode (and "l3" mode) traffic through the ipvlan interface **only** traverses iptables rules in the container's network namespace, which causes all traffic to bypass kube-proxy iptables rules. This is fine for experiment traffic, but k8s traffic needs to pass through those rules for everything to work as expected. This PR sets to the ipvlan mode to "l3s" which is a special mode to allow for traffic to pass through iptables rules in the default network namespace.

Additionally, this PR introduces a new flannel NetworkAttachmentDefinition which will be applied to experiment DaemonSets. This new config sets to `isDefaultGateway=false` so that experiment traffic is routed through the ipvlan interface, and not the flannel bridge.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/k8s-support/166)

### DIFF
--- a/k8s/custom-resource-definitions/network-attachment-definitions.yml
+++ b/k8s/custom-resource-definitions/network-attachment-definitions.yml
@@ -28,6 +28,8 @@ spec:
 
 # The CRD objects.
 ---
+# The default flannel configuration that gets applied to non-experiment pods.
+# The important bit here is that isDefaultGateway=true.
 apiVersion: "k8s.cni.cncf.io/v1"
 kind: NetworkAttachmentDefinition
 metadata:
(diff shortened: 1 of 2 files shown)


k8s/custom-resource-definitions/network-attachment-definitions.yml
```diff
@@ -28,6 +28,8 @@ spec:
 
 # The CRD objects.
 ---
+# The default flannel configuration that gets applied to non-experiment pods.
+# The important bit here is that isDefaultGateway=true.
```
pboothe: "bit" -> "difference"

nkinkade: Done.

PR #178, opened 2019-04-16
Adds the nodeinfo daemonset to core

pboothe:
This change is [Reviewable](https://reviewable.io/reviews/m-lab/k8s-support/178)

### DIFF
--- a/k8s/daemonsets/core/nodeinfo.yml
+++ b/k8s/daemonsets/core/nodeinfo.yml
@@ -20,6 +20,10 @@ spec:
     spec:
       nodeSelector:
         mlab/type: 'platform'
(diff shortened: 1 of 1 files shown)


pboothe:
It is currently running in sandbox.

pboothe:
Solid observation. I don't think we can get around that right now - in order for `route` and `ifconfig` to have useful output we're going to want to keep `hostNetwork:true`.  But it is a good thing to point out in a comment in the k8s yml file.

nkinkade:
`kube-rbac-proxy`! Example usage:

https://github.com/m-lab/k8s-support/blob/master/k8s/daemonsets/core/node-exporter.yml#L69

PR #185, opened 2019-04-18
Nodeinfo with an external config

pboothe:
This change is [Reviewable](https://reviewable.io/reviews/m-lab/k8s-support/185)


pboothe:
Corrected names everywhere and added automatic deployment of the configmap.

PR #200, opened 2019-04-29
Sets the update strategy for reboot-api to Recreate

nkinkade:
Apparently `RollingUpdate` is the default, but you can't do a rolling update on a deployment with only a single replica. It will hang with the old pod still running and the new one pending.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/k8s-support/200)


pboothe:
I already integrated this change into my massive PR that I am submitting right now.

PR #218, opened 2019-06-04
Make an experiments library

pboothe:
Now that we have two experiments, we can factor out the things which they have in common and create a library.

This PR creates that library, and moves ndt and bismark to use that library.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/k8s-support/218)


pboothe:
PTAL

pboothe:
Part of #221

PR #226, opened 2019-07-02
Adds a new CLUO operator and agent workloads for platform nodes

nkinkade:
We currently Container Linux Update Operator operator and agent workloads for the master cluster, which uses a very conservative reboot window. We want to experiment with using CLUO as a mechanism for performing rolling reboots of platform nodes. This PR adds new CLUO operator and agent workloads for platform nodes. It also renames the operator and agent workloads for the master cluster to clearly distinguish purpose.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/k8s-support/226)


nkinkade:
I just realized that I haven't thought about how the two operators are going to distinguish between platform nodes and master nodes. This PR is currently incomplete.

pboothe:
All platform nodes are tagged differently from cloud nodes, and I believe this can have a `nodeselector` somewhere in it when deployed. So the `mlab/type=platform` node label should be added to the `nodeselector` field when deploying the update operator.

nkinkade:
The node selector is there for the update-agent DaemonSet (you'll see it in this PR), but I'm not yet aware of any way to tell an update-operator instance to only pay attention to nodes with certain labels. There may be a way, I just haven't found it yet.

robertodauria:
Looks like the inner workings of the update-operator are mostly undocumented, so what follows is inferred from the source code.

The `checkBeforeReboot` [func](https://github.com/coreos/container-linux-update-operator/blob/master/pkg/operator/operator.go#L375) considers all the nodes with the default (constant) `before-reboot=true` but also checks for a configurable list of annotations that [can be set as comma-separated list via the command line](https://github.com/coreos/container-linux-update-operator/blob/master/cmd/update-operator/main.go#L32) (`-before-reboot-annotations=...`) and must be present for a node to be rebooted. I believe having two separate instances of the reboot-operator checking for different annotations could be a solution. Does it make sense?

nkinkade:
The changes in the PR will successfully annotate platform and master nodes such that only one update-operator should be allowed to reboot each kind of node. However, I've just run into another issue which may be a complete blocker, and that is that apparently update-operator instances hold leader elections such that, presumably, not more than one instance of an update-operator can will ever be in control of update-agents.

nkinkade:
This PR has morphed over the past couple weeks. Previously, we expected that we could run multiple update-operator instances and have one operate on the master cluster, while the other would manage the platform nodes. This turned out not to be practical, largely because the update-operators implement a leader election if multiple instances exist, meaning that running two instances to manage two separate groups of nodes is not currently supported at all with CLUO, and would require code modifications to make work.

As a workaround, we decided to just reboot all master nodes every week, each node on a different day (one of three days, currently Tue, Wed, Thu). And now the update-operator is back to running on a master node, and the update-agents only run on platform nodes.

To implement master node reboots, this PR implements a new systemd service and associated timer on all master nodes. Each master node has a file at `/etc/reboot-node-day` with a short day-of-the-week specification. The systemd `reboot-node.timer` runs daily, calling `reboot-node.service` . The service reads in the `/etc/reboot-node-day` file, and if the current day is equal to the content of that file, it reboots the machine. In this way, if the `update-engine.service` on the master nodes has downloaded a newer version of CoreOS, then these scheduled reboots should bring the nodes up to date.

PR #235, opened 2019-07-23
Gets rid of the node_exporter synthetic textfile metrics

nkinkade:
Since we've had a platform cluster we have synthetically created a number of metrics for platform cluster nodes to avoid alerts for legacy metrics that don't exist on the platform cluster. Basically, we are currently faking metrics for `collectd`, `vdlimit_*` and `lame-duck` on the platform cluster to avoid alerts firing for these legacy metrics. This PR removes all that, and is dependent on PR https://github.com/m-lab/prometheus-support/pull/506. Once the following PR is merged, then this PR can be merged and we shouldn't have to fake any more metrics to avoid alerts.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/k8s-support/235)


nkinkade:
@pboothe: PR https://github.com/m-lab/prometheus-support/pull/506 was merged into master which unblocks proceeding with this PR.

pboothe:
Sorry I did not get to this in a timely manner. Thanks @robertodauria for stepping in!

PR #242, opened 2019-07-31
go/tgifConfigmaps now are write-once (except for Prometheus)

pboothe:
This change is [Reviewable](https://reviewable.io/reviews/m-lab/k8s-support/242)


pboothe:
This was created accidentally. I hope this does not cause a push

PR #247, opened 2019-08-07
Allow for sandbox-only k8s objects

nkinkade:
Currently, there is no way to pin a certain object (e.g., DaemonSet) to the mlab-sandbox platform cluster. As it is, once a change gets merged into master it is then automatically deployed to staging, and from there once the repository gets tagged it automatically goes to production. However, there are going to be cases where we want something merged to master to stay only in the sandbox cluster. As a case in point, we currently want the BISMark experiment to remain in sandbox indefinitely (until the developers test the container, or rewrite it altogether).

This PR attempts to resolve this situation by creating a new `system-sandbox.jsonnet` file which will contain all objects that should only be deployed to the sandbox platform cluster. This new file only gets evaluated and applied to the cluster if the project is `mlab-sandbox` else it is skipped.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/k8s-support/247)

### DIFF
--- a/manage-cluster/apply_k8s_configs.sh
+++ b/manage-cluster/apply_k8s_configs.sh
@@ -47,3 +47,11 @@ kubectl apply -f secret-configs/
 kubectl apply -f system.json || true
 kubectl apply -f system.json || true
 kubectl apply -f system.json
+
+# Apply sandbox-only configurations. As mentioned in the comment above, to be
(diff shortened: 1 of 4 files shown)


nkinkade:
Closing this PR. It was not well thought out, and is superseded by the vastly more simple PR #248, which was merged.

PR #263, opened 2019-08-16
Add "utilization" experiment to daemonsets/core (aka DISCO, collectd)

stephen-soltesz:
This change adds a new daemonset/core experiment "utilization". This experiment generates two datatypes from local "switch" data and local "system" monitoring.

The utilization experiment only monitors the local environment and the switch via SNMP. Because there are no TCP-based connections (other than from pusher) the TCPINFO and TRACEROUTE side-cars are not included.

The `Pusher` function in templates.jsonnet is now public so that experiments like utilization can use the existing template without adding unnecessary side-cars.

The utilization experiment needs `hostNetwork: true` in order to discover the eth0 device, it's MAC address and lookup that MAC address in the tables returned from the switch.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/k8s-support/263)

### DIFF
--- a/k8s/daemonsets/core/utilization.jsonnet
+++ b/k8s/daemonsets/core/utilization.jsonnet
@@ -77,11 +77,11 @@ local exp = import '../templates.jsonnet';
           },
           {
             name: 'utilization-data',
-		    hostPath: {
+            hostPath: {
               path: '/cache/data/utilization',
               type: 'DirectoryOrCreate',
             },
-		  },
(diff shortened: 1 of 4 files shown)


stephen-soltesz:
https://github.com/m-lab/dev-tracker/issues/269

stephen-soltesz:
Verified that pusher was uploading data from the right place and to the right place.

PR #267, opened 2019-08-22
Update NDT version to v0.12.0

stephen-soltesz:
This change is [Reviewable](https://reviewable.io/reviews/m-lab/k8s-support/267)


stephen-soltesz:
Verified output is compatible with new parser.

static/widget client appears to WAI. the download test froze a few times -- but I'm not sure if that's my system or the client or the server.

Issue #268, opened 2019-08-27
Implement production canary releases


nkinkade:
We currently only have sandbox -> staging -> production. staging, for now, only get 0.001 of production traffic, which isn't really enough to compare a staging release with an existing production node to be sure that a deployment is working as expected. We need to have a true production canary release cycle where changes made to staging are deployed to some subset of production machines for a period.

nkinkade:
It turns out that implementing this by using versioned DaemonSets is going to require a good deal more thought. Implementing versioned DaemonSets is pretty easy, as PR #276 demonstrates, but that PR does not take into account deployment details.  Specifically these questions need to be answered, and perhaps even others:

* How do we properly version any experiment DaemonSet that we wanted versioned? Relying on a Github repository tag to version, for example, an NDT DaemonSet does not take into account that the repository may change and be tagged for all sorts of code changes that have _nothing_ to do with updating the NDT DaemonSet. Yet, the NDT DaemonSet would get a new version anyway, and would require a redeployment.

* How do controlled deployments happen when deploying becomes a matter of tagging nodes, not relying on the built-in k8s RollingUpdate mechanisms. It would likely either require some hackish script that an operator would run, or perhaps some service we deploy to the cluster that would manage labeling nodes to do a controlled and incremental rollout. The former is not what we want, and the latter is yet to be written.

nkinkade:
The thing to do in the short term may be to implement something along the lines of [our original idea of just having two DaemonSets for NDT, a production one and a canary one](https://docs.google.com/document/d/1Aq_yw8DOnDWAltVvx8EI_KL5mGoqe2DaJPJwzlxXJ08/edit#heading=h.dx0iddst66ld). But even this needs more consideration to avoid some of the same pitfalls as above.

nkinkade:
For now, it has been decided to just implement this by directing more production traffic to staging via manually tweaking the `ReverseProxyProbability` in mlab-ns after a merge to master. Once we are confident that the release is fine, we will reset `ReverseProxyProbability` and then release to production.

PR #276, opened 2019-09-11
Implements versioned experiments

nkinkade:
Currently, when you tag this repository (or merge to master), every single config is deployed to production (or staging) en masse. It is an all-or-nothing proposition. We'd like to start doing canary production releases. This PR is the first step in implementing versioning to experiments (starting with NDT).

This implementation creates two daemon sets for configured experiments named like `ndt-<version>`. Each version has a `nodeSelector` specific to that version like `mlab/version: <version>`. Thus, the DaemonSet will not deploy to a node unless that node is labeled in that way. This means that tagging a repository no longer launches the NDT experiment, and that launching a new version to a node is now handled by (currently) an operator who labels nodes appropriately. In the future, labeling of nodes could be handled by some service we write, allowing for canaried, checkpointed, rolling releases. In the short term, an operator could label some subset of canary nodes with the new version, wait some period of time for validation, then relabel the rest of the fleet to completed the release. This also makes rolling back a release as simple as relabeling a node (as long as the DaemonSet for the release you want to rollback to still exists).


nkinkade:
This is intended to partly resolve #268.

pboothe:
We should just create a release and it should have associated node tags. Making two releases, like you are doing here, feels wrong. 

Also, it means that the process of moving from "canary" to "release" does not involve generating anything new, you just relabel nodes.  Basically, we can make this a lot simpler and that sounds like a good idea to me.

nkinkade:
@pboothe: How do you implement this with a single DaemonSet? The DaemonSet for the old release, with the old `nodeSelector`, needs to stay in place, so that pods keep running on the nodes with the old label. Maybe I'm misunderstanding something?

pboothe:
We need old daemonsets to stick around. Ugh.

This makes things tougher, because we want to both (a) have daemonsets not restart when nothing changes and (b) have daemonsets have versions.

What we wish could be true is to have one daemonset name, and have it deploy one way on nodes with one tag and another way (which is possibly the same way) on nodes with another tag. We want the deployment process to NOT restart things if nothing has changed between versions.

pboothe:
This is not the right thing right now. It's a good idea, but not for implementing until we have a better handle on a lot of issues.

PR #277, opened 2019-09-13
Add DASH experiment

robertodauria:
This PR adds a new DaemonSet for the DASH experiment - subset of neubot. The docker image has been built from [neubot/dash:v0.4.0](https://github.com/neubot/dash/tree/v0.4.0).

The experiment name is still `neubot`, but the datatype it generates is named `dash`.

TLS is configured, but it will not work until we change the experiment's FQDN to something like `neubot-mlabN-<site>.measurement-lab.org`. However, @bassosimone agrees there is no harm in adding TLS already with an incorrect certificate since DASH previously only worked with HTTP, and will avoid an additional PR in future when we change the FQDN.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/k8s-support/277)


k8s/daemonsets/templates.jsonnet
```diff
@@ -314,4 +315,7 @@ local Experiment(name, index, bucket, datatypes=[]) = ExperimentNoIndex(name, da
 
   // The NDT tag to use for containers.
   ndtVersion: ndtVersion,
+
+  // The DASH tag to use for containers.
+  dashVersion: dashVersion,
```
pboothe: No need to put this here, just put the version directly in the daemonset.

The NDT version is here because it is used in both ndt and ndtcloud dameonsets.

PR #278, opened 2019-09-13
Creates canary DaemonSets deployed to machines with mlab/release=canary

nkinkade:
This PR is the first step in resolving #268. This PR will create `<experiment>-canary` DaemonSets which are configured to only deploy to nodes with an `mlab/release=canary` label. Conversely, the normal experiment DaemonSet is configured to _not_ deploy to nodes with that label.

The `cloudbuild-canary.yaml` is as-yet untested, as it relies on a GCB trigger in mlab-oti that looks for pushes to the `master` branch and then runs a build configured using this file.

NOTE: While not technically necessary, this PR will create the `<exp>-canary` DaemonSets in sandbox and staging, mostly for the purposes of possibly testing the canary mechanism without affecting production nodes.

NOTE #2: We need to be careful when deploying this (if we deploy it). When you label a node with `mlab/release=canary`, the normal DaemonSet will immediately start terminating, but this takes 150s. Meanwhile, the new canary DaemonSet is also immediately deployed. It will repeated fail until the normal one finishing terminating due to an IP conflict, and will expose us to the condition (bug?) where flannel does not reap private IPs of failed pods, and we risk depleting all private flannel IPs.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/k8s-support/278)

### DIFF
--- a/canaries.jsonnet
+++ b/canaries.jsonnet
@@ -1,0 +1,8 @@
+{
+  kind: 'List',
+  apiVersion: 'v1',
+  items: [
+    // Daemonsets
+    import 'k8s/daemonsets/experiments/ndt.jsonnet',
+  ],
+}
--- a/canaries.jsonnet
+++ b/canaries.jsonnet
@@ -3,6 +3,8 @@
   apiVersion: 'v1',
   items: [
     // Daemonsets
+    import 'k8s/daemonsets/core/host.jsonnet',
+    import 'k8s/daemonsets/experiments/bismark.jsonnet',
     import 'k8s/daemonsets/experiments/ndt.jsonnet',
   ],
 }
--- a/cloudbuild-canary.yaml
+++ b/cloudbuild-canary.yaml
@@ -1,0 +1,13 @@
+steps:
+
+# Fetch the KUBECONFIG file from GCS.
+- name: 'gcr.io/cloud-builders/gsutil'
+  args: ['cp', 'gs://k8s-support-$PROJECT_ID/admin.conf', '/workspace/admin.conf']
+
+# Push the configs.
+- name: 'gcr.io/cloud-builders/kubectl'
+  dir: '/workspace/manage-cluster'
+  entrypoint: '/workspace/manage-cluster/deploy_canary_k8s_configs.sh'
+  args:
+  - $PROJECT_ID
+  - /workspace/admin.conf
--- a/k8s/daemonsets/experiments/ndt.jsonnet
+++ b/k8s/daemonsets/experiments/ndt.jsonnet
@@ -1,33 +1,10 @@
 local exp = import '../templates.jsonnet';
 local project = std.extVar('PROJECT_ID');
-local canary = std.extVar('CANARY');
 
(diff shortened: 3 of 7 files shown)


pboothe:
Could you instead make both of them have a positive node selector?  `mlab/release=production` and `mlab/release=canary`.  In that way, the removal of one daemonset and the addition of the next can be done independently.

nkinkade:
@pboothe: I was attempting to avoid having to relabel every node on the platform, especially since this is going to only affect 10 nodes a single time. In using `mlab/release=production`, are thinking of that as a way to evict the 10 NDT pods (on the 10 canary nodes) in advance? If that is the main reasoning, then it would be easy enough to simply taint those 10 nodes in advance with `NoExecute`, do the deployment to production, then remove the taint.

I was considering all this a bit more last night, and came to the conclusion that even this PR is likely the wrong approach. For one, it only canaries a few select experiments, not our entire configuration. It's conceivable that one of these experiments could depend on some other k8s config that will not be present in the production cluster. I don't think that is the case now, but it could be.

I began to feel last night that perhaps a better approach would be to just join the 10 canary machines to the staging cluster, but still have mlab-ns direct production traffic at them.

pboothe:
I'm starting to get wrapped around an axle here:

1. What are the canaries testing that staging does not?
2. What is staging testing that the canaries do not?
3. How scary are the non-overlapping pieces of that Venn Diagram?

nkinkade:
@pboothe:

1. They will be testing a staging deployment on real production nodes with normal, real production traffic.
2. Staging and the canaries are testing the same thing. The difference being that the canaries will be getting real production traffic to better validate a staging deployment.
3. The non-overlapping part is real production traffic. Everybody seems to agree that sending production traffic to a subset of canaries is much better than deploying everything at once and hoping nothing is wrong with the staging release that went unnoticed. The general concern is that staging gets little to not traffic. One thing I proposed in the past was that sometime after a staging release is updated, we simply bump up the ReverseProxyProbability for however much time we think we need.

pboothe:
If that is all the canaries are for, then I think that upping the ReverseProxyProbability is sufficient to get what we want. 

I am pretty sure that @stephen-soltesz had a larger purpose in mind, centered around the idea of providing a method for phased rollouts and easy rollbacks.

nkinkade:
@pboothe: What else did you image canaries were for? I'm surprised that there is confusion about this, especially after the long discussion you, myself and Stephen had about it. Am I misconstruing what the intention the intention of #268?

Are you suggesting that we shelf this whole idea until Stephen returns from vacation? I need to determine whether I should be spending any time on this any longer.

nkinkade:
Closing this PR. Upon further consideration, just turning up the `ReverseProxyProbability` in mlab-ns (24h?) prior to a release is a much simpler and robust solution to the whole issue of canarying a release.

Having a subset of perpetual canary nodes in production is awkward, and also takes a staging DaemonSet and separates it from the rest of the staging cluster and drops it in the current production cluster. This assumes that the DaemonSets are essentially hermetic and have no dependencies on other cluster configurations.

pboothe:
After discussing: canaries inside Google means "staged rollouts with defined pauses". It does not uniformly mean that outside of Google.

PR #330, opened 2019-12-06
Return traceroute to polling ss to prevent segfault frequency

stephen-soltesz:
This change is [Reviewable](https://reviewable.io/reviews/m-lab/k8s-support/330)


stephen-soltesz:
Manual inspection of dmesg logs indicate no segfaults on lga0t nodes 2,3,4 over last 1.5hrs.

@yachang can you confirm that the traceroute containers are WAI?

PR #335, opened 2019-12-10
Add SOCATProxy as standard part of core service deployments

stephen-soltesz:
This change adds a new sidecar container for core services so that we can easily connect to the metric and instrumentation ports using `kubectl port-forward`.

With this change an operator can:

```
$ # identify the pod to connect to somehow.
$ kubectl port-forward pod/<pod-name> <local-port>:<remote-port>
$ # access localhost:<local-port> using your browser, `go tool pprof <url>`, lynx, or other tool.
```

Note: kube-rbac-proxy allows unauthenticated access to localhost automatically. So, the above steps work for `hostNetwork:true` pods using the RBACProxy already.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/k8s-support/335)


stephen-soltesz:
https://github.com/m-lab/dev-tracker/issues/507

PR #360, opened 2020-01-23
Properly mount the wehe data directory

pboothe:
This change is [Reviewable](https://reviewable.io/reviews/m-lab/k8s-support/360)


pboothe:
Suffers from https://github.com/m-lab/pusher/issues/84 , but it works.

PR #388, opened 2020-03-09
Add SOCAT proxy for ndt-server

stephen-soltesz:
Currently SOCAT Proxy is deployed automatically for all core services that run as side-cars with experiments. This change adds the SOCAT proxy to the ndt-server as well. This change will help investigate memory usage spikes in the NDT server in response to bursts of client connections. We think the memory used is disproportionate to what should be required.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/k8s-support/388)


stephen-soltesz:
https://github.com/m-lab/dev-tracker/issues/559

PR #400, opened 2020-03-18
First revtr attempt

pboothe:
This change is [Reviewable](https://reviewable.io/reviews/m-lab/k8s-support/400)


pboothe:
First version seems to work. We should not deploy to prod until all privacy concerns are addressed.

PR #417, opened 2020-04-20
Bootstraps master cluster and virtual nodes using Ubuntu instead of CoreOS

nkinkade:
This PR represents the set of changes I found necessary to update our bootstrap scripts to build a master cluster and virtual nodes on top of Ubuntu instead of CoreOS.

Riding on the coattails of this PR is an upgrade to k8s from 1.15.10 to 1.16.7. 1.18.x is already GA, so this helps keep us from getting too far behind.  In addition to k8s upgrades, this PR upgrades some other version (CNI plugins, flannel, etc.). The upgrade to a newer version of cert-manager required a few changes to the way that was configured, represented in this PR.

Conveniently, the Ubuntu minimal images already contain `cloud-init`, so modifying the cloud-config.yaml files was fairly easy... just converting the coreos-specific things to generic cloud-init things.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/k8s-support/417)


nkinkade:
@stephen-soltesz: Another idea I was considering yesterday was a script that would download the admin.conf file for each project, naming them like `$HOME/.kube/kubeconfig.<project> then the script would setup 3 bash functions in the user's .bashrc something like:

```
function kubectl-sandbox {
    kubectl --kubeconfig="$HOME/.kube/kubeconfig.mlab-sandbox" $@
}
```
This way you don't have to be in some particular directory to run any command you want.

What do you think?

PR #426, opened 2020-04-27
Add locate verify key to ndt-server

stephen-soltesz:
This change updates the ndt-server version to v0.16.0 and enables verification of access tokens provided by the locate service (i.e. from monitoring or early locate service).

This change adds a new secret that contains the verify keys that the ndt-server loads on start. At this time, access tokens remain optional. However, if an access token is provided it must be valid.

Note this change also moves the gsutil copy of `setup_k8s.sh` from `create_k8s_configs.sh` to `apply_k8s_configs.sh` so that the create script does not have side effects.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/k8s-support/426)


stephen-soltesz:
https://github.com/m-lab/dev-tracker/issues/574

PR #450, opened 2020-06-25
Add new Deployment Reloader

nkinkade:
[Reloader](https://github.com/stakater/Reloader) will trigger a rolling update of, currently, ndt and neubot in the event that cert-manager updates the Secret containing the wildcard TLS certificate. This PR is intended to resolve issue https://github.com/m-lab/ops-tracker/issues/1117.

This has been tested to work in the sandbox cluster by changing the ClusterIssuer of the M-Lab TLS Certificate to "letsencypt-staging", which triggered an update, and both ndt and neubot were rolled out.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/k8s-support/450)


k8s/deployments/reloader.jsonnet
```diff
@@ -0,0 +1,56 @@
...
+        },
+        labels: {
+          workload: 'reloader',
+        },
+      },
+      spec: {
+        containers: [
+          {
+            args: [
+              '--resources-to-ignore=configMaps',
```
pboothe, line 29: I am so glad we have a different solution for configmaps.

PR #481, opened 2020-08-12
Max timeout and new envelope version

stephen-soltesz:
This change updates the envelope version to v0.0.2 which adds dualstack support to the envelope and sets a max timeout of 10m to allow very long measurement runs by wehe clients.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/k8s-support/481)


stephen-soltesz:
I'm realizing that the dual stack support is moot given that `-envelope.token-required=false` -- which disables the ip management. but it should be ready when wehe is.

PR #483, opened 2020-08-18
Replace Fluentd with Vector

robertodauria:
This PR removes all the Fluentd resources and installs Vector via Helm.


robertodauria (commented):
> *[Reviewable](https://reviewable.io/reviews/m-lab/k8s-support/483)* status: 0 of 1 LGTMs obtained

___
*[manage-cluster/apply_k8s_configs.sh, line 74 at r1](https://reviewable.io/reviews/m-lab/k8s-support/483#-MF2W5KYCVNbl5Fgydiv:-MF4wN9bByODIQ9tLmtq:b-896fix) ([raw file](https://github.com/m-lab/k8s-support/blob/4c8e4b78922435d92576afaa517da1b46e104674/manage-cluster/apply_k8s_configs.sh#L74)):*

Instead of modifying the file in place, could we renamed `config/vector/values.yaml` to `values.yaml.template` and then have then redirect the `sed` output to `values.yaml`? I find it helpful for templates to have a suffix of `.template` so you know what it is and how it will be used without ever having to open it.
</blockquote></details>

Done.
___
*[manage-cluster/apply_k8s_configs.sh, line 77 at r1](https://reviewable.io/reviews/m-lab/k8s-support/483#-MF2WVvR3gb22mwtCz9-:-MF4w8a7-TI65DwPpwSO:b-1wfpz) ([raw file](https://github.com/m-lab/k8s-support/blob/4c8e4b78922435d92576afaa517da1b46e104674/manage-cluster/apply_k8s_configs.sh#L77)):*

Is a nightly version what we really want?
</blockquote></details>

Done. Nightly versions are all it's available for this Helm chart as they have just added it.I've added a TODO for myself to update it as soon as they release a new Vector version and generate the corresponding stable chart.
___
*[config/vector/values.yaml, line 3 at r1](https://reviewable.io/reviews/m-lab/k8s-support/483#-MF2VTjh6-gncEJVydQM:-MF4vzu59CEA_j-e3MiU:bo8fmkm) ([raw file](https://github.com/m-lab/k8s-support/blob/4c8e4b78922435d92576afaa517da1b46e104674/config/vector/values.yaml#L3)):*

How hard would it be change references to `fluent` to `vector`, or similar. These credentials will work for any tool, but it would be more intuitive if the naming reflected Vector instead of the old tool. Same for the volume and volume mount.
</blockquote></details>

Done. The service account has been created and uploaded to GCS for each project.

PR #495, opened 2020-09-22
Adds a `site` label to all metrics exported by disco

nkinkade:
I chose to implement this as a `relabel_config` as opposed to baking the label into disco directly. That may not be right, but it was a bit easier and seemed more like a post-processing task and not something that necessarily should be native to disco. It will be easier to change in the future, should like unlikely necessity ever present itself.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/k8s-support/495)

### DIFF
--- a/config/prometheus/prometheus.yml
+++ b/config/prometheus/prometheus.yml
@@ -135,9 +135,9 @@ scrape_configs:
       # metrics from each machine in a pod, and hence there is not site label
       # by default, but rather only a machine label, from which we can extract
       # the site.
-      - source_labels: [container, ifAlias, machine]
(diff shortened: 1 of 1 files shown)


stephen-soltesz (approved):
:lgtm: -- this looks like the right thing to do to me -- I agree that DISCO is not the right place. I wouldn't object in the future to including the site label on more pod metrics, e.g. ndt-server. I've had to group by site or metro the "hard way" in prometheus queries.

> *[Reviewable](https://reviewable.io/reviews/m-lab/k8s-support/495#-:-MHriRr94g5b1w82vtS7:bw6rn98)* status: :shipit: complete! 1 of 1 LGTMs obtained

PR #497, opened 2020-09-23
Add recording rules for the NDT:EarlyWarning dashboard

nkinkade:
Now that DISCOv2 is running in production in the platform cluster, SNMP metrics are located in that cluster, not the prometheus-federation cluster. This PR copies all of the Prometheus recording rules used on the NDT:EarlyWarning dashboard to the platform cluster instance of Prometheus. They are copied verbatim from the prometheus-support repository, with the only modifications being a few changes to the indentation of comments.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/k8s-support/497)

### DIFF
--- a/config/prometheus/rules.yml
+++ b/config/prometheus/rules.yml
@@ -251,7 +251,7 @@ groups:
   - record: machine:node_filesystem_used_ratio_12h_prediction:90th_quantile_30m
     expr: quantile_over_time(0.9, machine:node_filesystem_ratio_bytes:predict_linear3h_12h[30m])
 
-	# 90th percentiles @ 1h.
+  # 90th percentiles @ 1h.
   - record: candidate_site:uplink:90th_quantile_1h
     expr: quantile_over_time(0.9, switch:ifHCOutOctets:bps2m{ifAlias="uplink", machine=~"mlab2.*"}[1h])
   # k8s
@@ -264,7 +264,7 @@ groups:
   - record: machine:node_filesystem_used_ratio_12h_prediction:90th_quantile_1h
     expr: quantile_over_time(0.9, machine:node_filesystem_used_ratio:predict_linear3h_12h[1h])
 
-	# 90th percentiles @ 2h.
+  # 90th percentiles @ 2h.
   - record: candidate_site:uplink:90th_quantile_2h
     expr: quantile_over_time(0.9, switch:ifHCOutOctets:bps2m{ifAlias="uplink", machine=~"mlab2.*"}[2h])
   # k8s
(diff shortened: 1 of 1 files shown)


stephen-soltesz (changes requested):
> *[Reviewable](https://reviewable.io/reviews/m-lab/k8s-support/497)* status: 0 of 1 LGTMs obtained

___
*[config/prometheus/rules.yml, line 202 at r1](https://reviewable.io/reviews/m-lab/k8s-support/497#-MHx8gAb6u_6yHm5EYal:-MHx8gAb6u_6yHm5EYam:b-y36iwq) ([raw file](https://github.com/m-lab/k8s-support/blob/904f992485659a85274f45fe2fc85f1cf37a0fe8/config/prometheus/rules.yml#L202)):*
> ```YAML
>   # Units: bits per second.
>   - record: switch:ifHCOutOctets:bps2m
>     expr: 8 * irate(ifHCOutOctets{ifAlias="uplink", machine=~"mlab2.*"}[4m])
> ```

is there a way to do this using 'max' by site or similar rather than choosing one node (which could easily be down)? Or, run the recording rule for all machines and select the 'max' later in the dashboards?

Picking one node feels too fragile. What alternatives did you consider?

nkinkade (commented):
> *[Reviewable](https://reviewable.io/reviews/m-lab/k8s-support/497)* status: 0 of 1 LGTMs obtained

___
*[config/prometheus/rules.yml, line 202 at r1](https://reviewable.io/reviews/m-lab/k8s-support/497#-MHx8gAb6u_6yHm5EYal:-MI0pBpvESfKBVI8thEb:b-anascj) ([raw file](https://github.com/m-lab/k8s-support/blob/904f992485659a85274f45fe2fc85f1cf37a0fe8/config/prometheus/rules.yml#L202)):*

is there a way to do this using 'max' by site or similar rather than choosing one node (which could easily be down)? Or, run the recording rule for all machines and select the 'max' later in the dashboards?

Picking one node feels too fragile. What alternatives did you consider?
</blockquote></details>

Good point. I've modified the query to use `max by (site)` for switch metrics. I've also removed the redundant `ifAlias="uplink"` filter from all recording rules using the `switch:ifHCOutOctets:bps2m` recording rule, since that rule itself already filtered on `ifAlias="uplink"`.

stephen-soltesz (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/m-lab/k8s-support/497#-:-MI0prpQ-T0MPeYGDwEB:bnfp4nl)* status: :shipit: complete! 1 of 1 LGTMs obtained

PR #508, opened 2020-10-26
Create GCP resources for the new bmc-store-password ePoxy extension service

nkinkade:
This PR is part of https://github.com/m-lab/epoxy/issues/94. 

We are deploying a new ePoxy extension service running on each api-server. The extension will store node BMC passwords in GCD. This PR updates our platform bootstrap scripts to create all the necessary GCP resources to allow the extension to function properly. 

A quick way of summarizing what was done in this PR: copy all of the pieces that configured support for the token-server ePoxy extension and rename them with "bmc-store-password". Additionally, because the token-server was previously the only extension, this PR modifies the names of some token-server-related variables to distinguish them from the bmc-store-password variables.

Additionally, this PR updates the Go Docker image to 1.15 for building gcsfuse, which apparently no longer likes Go 1.13.

These changes were tested by tearing down the sandbox cluster and rebuilding it. Everything deployed fine, and it *appears* that all of the necessary GCP resources were created as expected.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/k8s-support/508)

### DIFF
--- a/manage-cluster/add_k8s_master_node.sh
+++ b/manage-cluster/add_k8s_master_node.sh
@@ -95,7 +95,8 @@ gcloud compute ssh "${BOOTSTRAP_MASTER}" "${GCP_ARGS[@]}" --zone "${BOOTSTRAP_MA
 EOF
 
 # If they exist, delete the node name from various loadbalancer group resources.
-delete_token_server_backend "${GCE_NAME}" "${GCE_ZONE}"
+delete_server_backend "${GCE_NAME}" "${GCE_ZONE}" "${TOKEN_SERVER_BASE_NAME}"
+delete_server_backend "${GCE_NAME}" "${GCE_ZONE}" "${BMC_STORE_PASSWORD_BASE_NAME}"
 delete_target_pool_instance "${GCE_NAME}" "${GCE_ZONE}"
 delete_instance_group "${GCE_NAME}" "${GCE_ZONE}"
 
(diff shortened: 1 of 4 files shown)


stephen-soltesz (approved):
:lgtm: -- there are a lot of pieces.. it's still unfortunate that we can't manage our cluster with gke -- these extensions could be deployed in a gke cluster very easily... maybe epoxy too.

> *[Reviewable](https://reviewable.io/reviews/m-lab/k8s-support/508#-:-MKekF9dAL4m-oIroEY_:bizk77r)* status: :shipit: complete! 1 of 1 LGTMs obtained

nkinkade (commented):
There are a lot of pieces, and the fact that it is all done with shell scripts makes it even more convoluted-seeming. I was wondering yesterday if this could somehow be made cleaner if we made all the API calls directly rather than through `gcloud`. I don't know what this would look like, but I was imagining a Go utility that read in a list of API calls  and operated on them sequentially. I was then imagining that the "config" could be a JSON file, or even just a text file with all the API calls. It may be a bad idea and not even possible or workable, but I was dismayed yesterday at how unwieldy the shell scripts are.

> *[Reviewable](https://reviewable.io/reviews/m-lab/k8s-support/508#-:-MKepLxi2-S5j5xjWTDF:b-2htx9e)* status: :shipit: complete! 1 of 1 LGTMs obtained

PR #513, opened 2020-11-09
Adds recording rule for container restarts.

nkinkade:
This change is [Reviewable](https://reviewable.io/reviews/m-lab/k8s-support/513)

### DIFF
--- a/config/prometheus/rules.yml
+++ b/config/prometheus/rules.yml
@@ -185,8 +185,8 @@ groups:
   # The platform cluster has tens of thousands of containers, which makes
   # calculating container restarts very expensive, even prohibitive over a long
   # enough period. This recording rule should help speed things up.
-  - record: container:kube_pod_container_status_restarts:sum1d
-    exp: (kube_pod_container_status_restarts_total - (kube_pod_container_status_restarts_total offset 1d)) > 0
+  - record: container:kube_pod_container_status_restarts:increase1d
+    expr: (kube_pod_container_status_restarts_total - (kube_pod_container_status_restarts_total offset 1d)) > 0
 
 ## NDT Early Warning aggregation rules.
 #
--- a/config/prometheus/rules.yml
+++ b/config/prometheus/rules.yml
@@ -185,8 +185,8 @@ groups:
   # The platform cluster has tens of thousands of containers, which makes
   # calculating container restarts very expensive, even prohibitive over a long
   # enough period. This recording rule should help speed things up.
-  - record: container:kube_pod_container_status_restarts:increase7d
-    exp: increase(kube_pod_container_status_restarts_total[7d])
+  - record: container:kube_pod_container_status_restarts:sum1d
+    exp: (kube_pod_container_status_restarts_total - (kube_pod_container_status_restarts_total offset 1d)) > 0
 
 ## NDT Early Warning aggregation rules.
 #
--- a/config/prometheus/rules.yml
+++ b/config/prometheus/rules.yml
@@ -176,11 +176,17 @@ groups:
   - record: datatype:pusher_bytes_per_tarfile:increase24h
     expr: sum by(datatype) (increase(pusher_bytes_per_tarfile_sum[1d]))
 
-  ## Ops: Tactical & SRE Overview Dashboard.
-  #
+## Ops: Tactical & SRE Overview Dashboard.
+
   # This rule speeds up the probably OOM query used by the SRE overview dashboard.
   - record: err:dmesg_logs:increase_24h
     expr: increase(dmesg_logs{priority="err"}[24h]) > 0
+
+  # The platform cluster has tens of thousands of containers, which makes
+  # calculating container restarts very expensive, even prohibitive over a long
+  # enough period. This recording rule should help speed things up.
+  - record: container:kube_pod_container_status_restarts:increase7d
+    exp: increase(kube_pod_container_status_restarts_total[7d])
 
 ## NDT Early Warning aggregation rules.
 #


stephen-soltesz (changes requested):
> *[Reviewable](https://reviewable.io/reviews/m-lab/k8s-support/513)* status: 0 of 1 LGTMs obtained

___
*[config/prometheus/rules.yml, line 189 at r1](https://reviewable.io/reviews/m-lab/k8s-support/513#-MLjFUA73ILIZixHxljg:-MLjFUA73ILIZixHxljh:b-bdugz2) ([raw file](https://github.com/m-lab/k8s-support/blob/3ad807099c49346b9832fd0a302acbe09eec550e/config/prometheus/rules.yml#L189)):*
> ```
> increase(kube_pod_container_status_restarts_total[7d])
> ```

Hmm... in mlab-oti this produced the error:

```
Error executing query: query processing would load too many samples into memory in query execution
```

There are 29151 timeseries with that metric.. this RR would double it..

Thinking out loud: could we do increase over a shorter period of time (e.g. 1day), and then run a `sum_over_time()` in either a new recording rule or the SRE dashboard? We also need the query time to be relatively low for the RR to be efficient. It will be running every 60sec.

Happy to talk through this more for additional ideas.

stephen-soltesz (changes requested):
> *[Reviewable](https://reviewable.io/reviews/m-lab/k8s-support/513)* status: 0 of 1 LGTMs obtained

___
*[config/prometheus/rules.yml, line 189 at r1](https://reviewable.io/reviews/m-lab/k8s-support/513#-MLjFUA73ILIZixHxljg:-MLjICr60WCNfu6-HuqE:b-6mwr34) ([raw file](https://github.com/m-lab/k8s-support/blob/3ad807099c49346b9832fd0a302acbe09eec550e/config/prometheus/rules.yml#L189)):*

> ```
> increase(kube_pod_container_status_restarts_total[7d])
> ```

Hmm... in mlab-oti this produced the error:

```
Error executing query: query processing would load too many samples into memory in query execution
```

There are 29151 timeseries with that metric.. this RR would double it..

Thinking out loud: could we do increase over a shorter period of time (e.g. 1day), and then run a `sum_over_time()` in either a new recording rule or the SRE dashboard? We also need the query time to be relatively low for the RR to be efficient. It will be running every 60sec.

Happy to talk through this more for additional ideas.
</blockquote></details>

Another idea, this seems to run relatively quickly:

```
(kube_pod_container_status_restarts_total - (kube_pod_container_status_restarts_total offset 1d)) > 0
```

This is all restarts that are increasing over the last day.

nkinkade (commented):
> *[Reviewable](https://reviewable.io/reviews/m-lab/k8s-support/513)* status: 0 of 1 LGTMs obtained

___
*[config/prometheus/rules.yml, line 189 at r1](https://reviewable.io/reviews/m-lab/k8s-support/513#-MLjFUA73ILIZixHxljg:-MLjIP8C-ZUwcjk7st9G:bn8ybny) ([raw file](https://github.com/m-lab/k8s-support/blob/3ad807099c49346b9832fd0a302acbe09eec550e/config/prometheus/rules.yml#L189)):*

Another idea, this seems to run relatively quickly:

```
(kube_pod_container_status_restarts_total - (kube_pod_container_status_restarts_total offset 1d)) > 0
```

This is all restarts that are increasing over the last day.
</blockquote></details>

Yeah, the error you got in mlab-oti was the motivation for this PR. The "container restart" panel in the existing SRE dashboard is already very slow.  I seems that I didn't thoroughly think this recording rule through very well. 

I just now saw your updated comment about a query that runs relatively quickly. I came back to this PR to post another possible solution, which is not dissimilar to your notion above. Funny that we were working on the same basic notion at the same time without knowing it:

```
(kube_pod_container_status_restarts_total - kube_pod_container_status_restarts_total offset 7d) > 10
```

The above query does take around 8-10s to run, which is still pretty slow. But we mostly only use the SRE Overview dashboard once a week, and a 10s load time isn't that onerous.

stephen-soltesz (changes requested):
> *[Reviewable](https://reviewable.io/reviews/m-lab/k8s-support/513)* status: 0 of 1 LGTMs obtained

___
*[config/prometheus/rules.yml, line 189 at r1](https://reviewable.io/reviews/m-lab/k8s-support/513#-MLjFUA73ILIZixHxljg:-MM1NsWH2Mct4mmomntZ:bowqp9r) ([raw file](https://github.com/m-lab/k8s-support/blob/3ad807099c49346b9832fd0a302acbe09eec550e/config/prometheus/rules.yml#L189)):*

Yeah, the error you got in mlab-oti was the motivation for this PR. The "container restart" panel in the existing SRE dashboard is already very slow.  I seems that I didn't thoroughly think this recording rule through very well. 

I just now saw your updated comment about a query that runs relatively quickly. I came back to this PR to post another possible solution, which is not dissimilar to your notion above. Funny that we were working on the same basic notion at the same time without knowing it:

```
(kube_pod_container_status_restarts_total - kube_pod_container_status_restarts_total offset 7d) > 10
```

The above query does take around 8-10s to run, which is still pretty slow. But we mostly only use the SRE Overview dashboard once a week, and a 10s load time isn't that onerous.
</blockquote></details>

Okay, yeah, your query works too for the recording rule. Using 7d would mean that most of the values on the dashboard would include periods from the previous week. Using 1d would limit that to the first day (of a 7d panel).

It doesn't take 10s to load for 7d panel, right? We still want the RR afaict.

nkinkade (commented):
@stephen-soltesz: PTAL? I've updated the recording rule with your suggestion, to sum container restarts over 1d. That is more flexible than 7d for the cases where we may want to calculate over less than 7d and should still be fast.

> *[Reviewable](https://reviewable.io/reviews/m-lab/k8s-support/513#-:-MM1gH1F5QUP3IWVJmEz:b-3m6jt)* status: 0 of 1 LGTMs obtained

stephen-soltesz (approved):
:lgtm: - with name change and typo fix.

> *[Reviewable](https://reviewable.io/reviews/m-lab/k8s-support/513#-:-MM2edU5BZm7hprCMOtJ:b-472juv)* status: :shipit: complete! 1 of 1 LGTMs obtained

___
*[config/prometheus/rules.yml, line 188 at r2](https://reviewable.io/reviews/m-lab/k8s-support/513#-MM2eU_029zUorZOLu6x:-MM2eU_029zUorZOLu6y:b-eob75v) ([raw file](https://github.com/m-lab/k8s-support/blob/f77b0d11b0c5e6001149fb91ff1add57e4ba9a86/config/prometheus/rules.yml#L188)):*
> ```YAML
>   # calculating container restarts very expensive, even prohibitive over a long
>   # enough period. This recording rule should help speed things up.
>   - record: container:kube_pod_container_status_restarts:sum1d
> ```

The operation suffix is not `sum`. How about `increase1d` ?

It might be worth while that the expression is a "lighter weight but lossy `increase()`".
___
*[config/prometheus/rules.yml, line 189 at r2](https://reviewable.io/reviews/m-lab/k8s-support/513#-MM2g08h3OYEq3JNGlsf:-MM2g08h3OYEq3JNGlsg:b-lu9wps) ([raw file](https://github.com/m-lab/k8s-support/blob/f77b0d11b0c5e6001149fb91ff1add57e4ba9a86/config/prometheus/rules.yml#L189)):*
> ```YAML
>   # enough period. This recording rule should help speed things up.
>   - record: container:kube_pod_container_status_restarts:sum1d
>     exp: (kube_pod_container_status_restarts_total - (kube_pod_container_status_restarts_total offset 1d)) > 0
> ```

typo: `exp` -> `expr`.

nkinkade (commented):
> *[Reviewable](https://reviewable.io/reviews/m-lab/k8s-support/513)* status: :shipit: complete! 1 of 1 LGTMs obtained

___
*[config/prometheus/rules.yml, line 188 at r2](https://reviewable.io/reviews/m-lab/k8s-support/513#-MM2eU_029zUorZOLu6x:-MM2mK2X2KJXa7k1Kbc9:byflx02) ([raw file](https://github.com/m-lab/k8s-support/blob/f77b0d11b0c5e6001149fb91ff1add57e4ba9a86/config/prometheus/rules.yml#L188)):*
I chose `sum` because [the Prometheus recommendation](https://prometheus.io/docs/practices/rules/#naming-and-aggregation) is to use that when there is no obvious operation:

> If there is no obvious operation to use, use sum. 

But sure, `increase` seems good, since the end value minus the start value is effectively the increase for the interval.
___
*[config/prometheus/rules.yml, line 189 at r2](https://reviewable.io/reviews/m-lab/k8s-support/513#-MM2g08h3OYEq3JNGlsf:-MM2nss70vKsOdI7C9cU:b-d5q2ho) ([raw file](https://github.com/m-lab/k8s-support/blob/f77b0d11b0c5e6001149fb91ff1add57e4ba9a86/config/prometheus/rules.yml#L189)):*

typo: `exp` -> `expr`.
</blockquote></details>

Fixed. Thanks.

PR #526, opened 2020-12-15
Small changes necessary to migrate CRI from Docker to containerd

nkinkade:
**cniVersion**
For reasons that I still don't understand, using `cniVersion: 0.2.0` in our main CNI config file does not work with containerd as the CRI. With this configuration, containerd complains:

```
failed to find network info
``` 

It was actually partly by chance that I even experimented with changing this value to `0.3.1`. That it works with this version I can only explain by presuming that multus _does_ support `0.3.1` (and probably `0.4.0`, too), and the only thing that doesn't is our IPAM plugin `index2ip`.  Likely, the CRI calls multus requesting cniVersion of 0.3.1, which is fine since multus supports this, and we still have our `network-attachment-definitions` [using a cniVersion of 0.2.0](https://github.com/m-lab/k8s-support/blob/master/k8s/networks/networks.jsonnet#L56), so multus may expect an 0.2.0 result from `index2ip`. And our [flannel config](https://github.com/m-lab/k8s-support/blob/master/k8s/networks/networks.jsonnet#L33) is already using `0.3.0`. I'm guessing that multus is taking the 0.2.0 output of `index2ip` and the `0.3.0` output of flannel and then somehow coalesces those into an `0.3.1` output for the CRI.

**cgroupDriver**
We need this to be systemd for the kubelet, since the containerd CRI (and runc OCI)  is also using this driver, and they need to match.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/k8s-support/526)

### DIFF
--- a/config/flannel.jsonnet
+++ b/config/flannel.jsonnet
@@ -37,10 +37,11 @@ local virtualCNIConf = {
 // It should not contain any index2ip stuff. It is the backup config for when
 // multus isn't working or a pod is not tagged with any network annotations.
 //
-// cniVersion:0.2.0 is the CNI version that index2ip knows about.
+// cniVersion:0.3.1 is a CNI version that multus knows about. The index2ip IPAM
+// plugin only knows about 0.2.0, but multus handles this.
 local physicalCNIConf = {
   name: 'multus-network',
-  cniVersion: '0.2.0',
+  cniVersion: '0.3.1',
   type: 'multus',
   kubeconfig: '/etc/kubernetes/kubelet.conf',
   multusNamespace: 'default',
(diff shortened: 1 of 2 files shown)


stephen-soltesz (approved):
This is hard core :+1: :lgtm:

Would you be interested in updating index2ip to be CNI v0.3 or v0.4 compatible? I'm imagining that it's primarily the output format rather than additional actions (but I don't know this).

> *[Reviewable](https://reviewable.io/reviews/m-lab/k8s-support/526#-:-MOckoYJ26RsJmSQPqY7:be5al6f)* status: :shipit: complete! 1 of 1 LGTMs obtained

nkinkade (commented):
There was already a very old open issue for this. I assigned it to myself and added to my current backlog of things to do. I think it makes sense to get that plugin updated:

https://github.com/m-lab/index2ip/issues/8

> *[Reviewable](https://reviewable.io/reviews/m-lab/k8s-support/526#-:-MOclqKQ3YFvUYcE4igy:bmau4j9)* status: :shipit: complete! 1 of 1 LGTMs obtained

PR #528, opened 2020-12-16
Migrate API cluster to using containerd instead of Docker for the CRI

nkinkade:
The changes will cause the bootstrap scripts to produce API nodes/clusters using containerd as the CRI instead of Docker.

In sandbox I was able to completely redeploy the cluster. In staging and production this can be accomplished by deleting and replacing one master node at a time and using the `k8s_add_master_node.sh` script to readd them incrementally.

Now that we are using a "remote" CRI, instead of Docker, the kubelet will take over watching container log sizes. There is one configuration in this PR to set the same log size limit we had with Docker.

**NOTE**: We are still using Docker on the api-server nodes to launch ePoxy extensions and the gcp-loadbalancer-proxy service.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/k8s-support/528)

### DIFF
--- a/manage-cluster/bootstraplib.sh
+++ b/manage-cluster/bootstraplib.sh
@@ -392,7 +392,7 @@ EOF
 	export ETCDCTL_CERT=/etc/kubernetes/pki/etcd/peer.crt
 	export ETCDCTL_KEY=/etc/kubernetes/pki/etcd/peer.key
 	export ETCDCTL_ENDPOINTS=https://127.0.0.1:2379
-  export CONTAINER_RUNTIME_ENDPOINT=unix:///run/containerd/containerd.sock
+	export CONTAINER_RUNTIME_ENDPOINT=unix:///run/containerd/containerd.sock
 EOF2
     ) | tee -a /root/.profile /root/.bashrc"
 EOF
--- a/manage-cluster/bootstraplib.sh
+++ b/manage-cluster/bootstraplib.sh
@@ -205,8 +205,10 @@ function create_master {
     # Configure containerd. We only need to make one change to the default
     # configuration, which is to configure runc to use the systemd cgroup
     # driver.
+    mkdir /etc/containerd
     containerd config default | \
-      sed -e '/containerd\.runtimes\.runc/a \          [plugins."io.containerd.grpc.v1.cri".containerd.runtimes.runc.options]\n            SystemdCgroup = true'
+      sed -e '/containerd\.runtimes\.runc/a \          [plugins."io.containerd.grpc.v1.cri".containerd.runtimes.runc.options]\n            SystemdCgroup = true' \
+      > /etc/containerd/config.toml
 
     # Enable and start the kubelet service
     systemctl enable --now kubelet.service
@@ -390,6 +392,7 @@ EOF
 	export ETCDCTL_CERT=/etc/kubernetes/pki/etcd/peer.crt
 	export ETCDCTL_KEY=/etc/kubernetes/pki/etcd/peer.key
 	export ETCDCTL_ENDPOINTS=https://127.0.0.1:2379
+  export CONTAINER_RUNTIME_ENDPOINT=unix:///run/containerd/containerd.sock
 EOF2
     ) | tee -a /root/.profile /root/.bashrc"
 EOF
--- a/manage-cluster/bootstraplib.sh
+++ b/manage-cluster/bootstraplib.sh
@@ -202,6 +202,12 @@ function create_master {
     cp etcd-${ETCDCTL_VERSION}-linux-amd64/etcdctl /opt/bin
     rm -rf etcd-${ETCDCTL_VERSION}-linux-amd64
 
+    # Configure containerd. We only need to make one change to the default
+    # configuration, which is to configure runc to use the systemd cgroup
+    # driver.
+    containerd config default | \
+      sed -e '/containerd\.runtimes\.runc/a \          [plugins."io.containerd.grpc.v1.cri".containerd.runtimes.runc.options]\n            SystemdCgroup = true'
+
     # Enable and start the kubelet service
     systemctl enable --now kubelet.service
     systemctl daemon-reload
--- a/manage-cluster/cloud-config_node.yml
+++ b/manage-cluster/cloud-config_node.yml
@@ -1,7 +1,7 @@
 #cloud-config
 
 packages:
-- docker.io
+- containerd
 - ebtables
 - iptables
 - socat
@@ -29,23 +29,21 @@ write_files:
   permissions: 0644
   owner: root:root
   content: |
+    overlay
     br_netfilter
 
-- path: /etc/docker/daemon.json
+# The odd number of 999 because GCE nodes already have a file named
+# 99-gce.conf, which, among other things, disables ip forwarding.
+- path: /etc/sysctl.d/999-containerd.conf
   permissions: 0644
   owner: root:root
   content: |
-    {
-      "exec-opts": [
-        "native.cgroupdriver=systemd"
-      ],
-      "log-driver": "json-file",
-      "log-opts": {
-        "max-size": "100m"
-      },
-      "storage-driver": "overlay2"
-    }
+    net.ipv4.ip_forward                 = 1
+    net.bridge.bridge-nf-call-iptables  = 1
+    net.bridge.bridge-nf-call-ip6tables = 1
 
 runcmd:
-- systemctl enable docker
-- systemctl start docker
+- sysctl --system
+- systemctl restart systemd-modules-load.service
+- systemctl enable containerd
+- systemctl start containerd
--- a/manage-cluster/kubeadm-config.yml.template
+++ b/manage-cluster/kubeadm-config.yml.template
@@ -21,6 +21,10 @@ controlPlane:
     bindPort: 6443
 nodeRegistration:
   name: "{{MASTER_NAME}}"
+  kubeletExtraArgs:
+    container-runtime: "remote"
+    container-runtime-endpoint: "/run/containerd/containerd.sock"
+    resolv-conf: "/run/systemd/resolve/resolv.conf"
 ---
 apiVersion: kubeadm.k8s.io/v1beta2
 kind: ClusterConfiguration
--- a/manage-cluster/kubeadm-config.yml.template
+++ b/manage-cluster/kubeadm-config.yml.template
@@ -2,6 +2,10 @@ apiVersion: kubeadm.k8s.io/v1beta2
 kind: InitConfiguration
 nodeRegistration:
   name: "{{MASTER_NAME}}"
+  kubeletExtraArgs:
+    container-runtime: "remote"
+    container-runtime-endpoint: "unix:///run/containerd/containerd.sock"
+    resolv-conf: "/run/systemd/resolve/resolv.conf"
 localAPIEndpoint:
   advertiseAddress: "{{INTERNAL_IP}}"
   bindPort: 6443
--- a/manage-cluster/kubeadm-config.yml.template
+++ b/manage-cluster/kubeadm-config.yml.template
@@ -23,7 +23,7 @@ nodeRegistration:
   name: "{{MASTER_NAME}}"
   kubeletExtraArgs:
     container-runtime: "remote"
-    container-runtime-endpoint: "/run/containerd/containerd.sock"
+    container-runtime-endpoint: "unix:///run/containerd/containerd.sock"
     resolv-conf: "/run/systemd/resolve/resolv.conf"
 ---
 apiVersion: kubeadm.k8s.io/v1beta2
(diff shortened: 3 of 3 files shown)


stephen-soltesz (approved):
Just general questions. Nothing blocking necessarily. :lgtm:

> *[Reviewable](https://reviewable.io/reviews/m-lab/k8s-support/528#-:-MOhq5Bz8WdMGef9yfUZ:b-6u9xna)* status: :shipit: complete! 1 of 1 LGTMs obtained

___
*[manage-cluster/bootstraplib.sh, line 210 at r1](https://reviewable.io/reviews/m-lab/k8s-support/528#-MOhpPDM1FxQDPIrZisk:-MOhpPDM1FxQDPIrZisl:b-j4uel5) ([raw file](https://github.com/m-lab/k8s-support/blob/49033e6c2e2b9d306c21fbde2cd96b0c291886ae/manage-cluster/bootstraplib.sh#L210)):*
> ```Shell
>     mkdir /etc/containerd
>     containerd config default | \
>       sed -e '/containerd\.runtimes\.runc/a \          [plugins."io.containerd.grpc.v1.cri".containerd.runtimes.runc.options]\n            SystemdCgroup = true' \
> ```

/me shakes fist at sky -- Why is there a new config format?

Not blocking, just thinking out loud -- this sed expression is a little hairy. is there a generic config updater command to add a new section and key/value ?
___
*[manage-cluster/kubeadm-config.yml.template, line 31 at r1](https://reviewable.io/reviews/m-lab/k8s-support/528#-MOhooyV7MNqy8E-T6MA:-MOhooyW9nJ_EzKHpykk:bkahd8j) ([raw file](https://github.com/m-lab/k8s-support/blob/49033e6c2e2b9d306c21fbde2cd96b0c291886ae/manage-cluster/kubeadm-config.yml.template#L31)):*
> ```
>     container-runtime: "remote"
>     container-runtime-endpoint: "unix:///run/containerd/containerd.sock"
>     resolv-conf: "/run/systemd/resolve/resolv.conf"
> ```

This reminded me of another us of resolv.conf -- is there a kubelet or containerd setting to remove the need for this? https://github.com/m-lab/k8s-support/blob/master/k8s/daemonsets/templates.jsonnet#L492
___
*[manage-cluster/kubeadm-config.yml.template, line 58 at r1](https://reviewable.io/reviews/m-lab/k8s-support/528#-MOhoU576_zwccMo5O3U:-MOhoU576_zwccMo5O3V:b7ddt1o) ([raw file](https://github.com/m-lab/k8s-support/blob/49033e6c2e2b9d306c21fbde2cd96b0c291886ae/manage-cluster/kubeadm-config.yml.template#L58)):*
> ```
> kind: KubeletConfiguration
> cgroupDriver: systemd
> containerLogMaxSize: 100Mi
> ```

This setting propagates to all nodes in the cluster? This is what is replacing the old docker logs config?

nkinkade (commented):
@stephen-soltesz: I made an additional commit (@4350049) yesterday evening, which you did not review here. It modifies the node cloud-init file to covert platform VMs (notably the Prometheus VM, currently) to use containerd instead of Dockder. 

PTAL?

> *[Reviewable](https://reviewable.io/reviews/m-lab/k8s-support/528#-:-MOldJd_9gPb_RXv39qc:b-xm08qp)* status: :shipit: complete! 1 of 1 LGTMs obtained

___
*[manage-cluster/bootstraplib.sh, line 210 at r1](https://reviewable.io/reviews/m-lab/k8s-support/528#-MOhpPDM1FxQDPIrZisk:-MOl_hYt9_3xnPM4LWnx:b-5sg82a) ([raw file](https://github.com/m-lab/k8s-support/blob/49033e6c2e2b9d306c21fbde2cd96b0c291886ae/manage-cluster/bootstraplib.sh#L210)):*

/me shakes fist at sky -- Why is there a new config format?

Not blocking, just thinking out loud -- this sed expression is a little hairy. is there a generic config updater command to add a new section and key/value ?
</blockquote></details>

I don't know of any tool or utility that can update the containerd config. 

When you say "new" config format, are you just grumbling about TOML? Which config format would make more sense to you?

I agree that having to use a `sed` statement in the first place not as pretty as one would like. The other alternative I considered was to create a static config file in our repo and then just upload that.  I opted against that because this seemed easier, since we are only adding 2 lines to the default config.

Personally, I don't find the `sed` statement to be hairy. It is a very basic  and common`sed` pattern. Maybe what makes it seem hairy is the string of text I'm inserting, which is rather long and tries to preserve formatting and indention in the resultant config file.
___
*[manage-cluster/kubeadm-config.yml.template, line 31 at r1](https://reviewable.io/reviews/m-lab/k8s-support/528#-MOhooyV7MNqy8E-T6MA:-MOlb7iJ7DIH_R-XocJ8:b-w04axp) ([raw file](https://github.com/m-lab/k8s-support/blob/49033e6c2e2b9d306c21fbde2cd96b0c291886ae/manage-cluster/kubeadm-config.yml.template#L31)):*

This reminded me of another us of resolv.conf -- is there a kubelet or containerd setting to remove the need for this? https://github.com/m-lab/k8s-support/blob/master/k8s/daemonsets/templates.jsonnet#L492
</blockquote></details>

I added this `resolv-conf` config here because I happened to observe in the kubelet logs a message stating that the flag `--resolv-conf` was deprecated and that the configuration should be moved to kubelet config file, which this does. I got the value by observing the value of the flag on a running kubelet (kubeadm sets the value).

I actually never knew what that `fix-resolv-conf` initContainer was for.  Someone else (@pboothe ?) created it, and I never really questioned it. Why do we need to much with that.  We can have this conversation elsewhere. I'm just mentioning here that I never knew why that was necessary.
___
*[manage-cluster/kubeadm-config.yml.template, line 58 at r1](https://reviewable.io/reviews/m-lab/k8s-support/528#-MOhoU576_zwccMo5O3U:-MOlc23m9WqNp29WavG7:bnqrsql) ([raw file](https://github.com/m-lab/k8s-support/blob/49033e6c2e2b9d306c21fbde2cd96b0c291886ae/manage-cluster/kubeadm-config.yml.template#L58)):*

This setting propagates to all nodes in the cluster? This is what is replacing the old docker logs config?
</blockquote></details>

That's right. The `KubeletConfiguration` part of the kubeadmin config ultimately gets stored in a ConfigMap in the cluster. When  a node joins the cluster using `kubeadm join`, `kubeadm` reads that ConfigMap and writes it out to the node at  `/var/lib/kubelet/config.yaml`, which the kubelet then consumes at runtime via the `--config` flag.

My understanding is that this config is what is replacing the old Docker configuration  of:

```
 "log-opts": {
    "max-size": "100m"
  },
```

I believe that when `--container-runtime=remote` that this setting becomes active, because Docker is not managing it.

stephen-soltesz (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/m-lab/k8s-support/528#-:-MOlggvR9yFSHO9hNE3E:bnfp4nl)* status: :shipit: complete! 1 of 1 LGTMs obtained

___
*[manage-cluster/bootstraplib.sh, line 210 at r1](https://reviewable.io/reviews/m-lab/k8s-support/528#-MOhpPDM1FxQDPIrZisk:-MOlfucr9KmcXWEeefr1:b-m64hw8) ([raw file](https://github.com/m-lab/k8s-support/blob/49033e6c2e2b9d306c21fbde2cd96b0c291886ae/manage-cluster/bootstraplib.sh#L210)):*

I don't know of any tool or utility that can update the containerd config. 

When you say "new" config format, are you just grumbling about TOML? Which config format would make more sense to you?

I agree that having to use a `sed` statement in the first place not as pretty as one would like. The other alternative I considered was to create a static config file in our repo and then just upload that.  I opted against that because this seemed easier, since we are only adding 2 lines to the default config.

Personally, I don't find the `sed` statement to be hairy. It is a very basic  and common`sed` pattern. Maybe what makes it seem hairy is the string of text I'm inserting, which is rather long and tries to preserve formatting and indention in the resultant config file.
</blockquote></details>

My "shakes fist" was mostly a joke. The part that wasn't a joke was directed at the maintainers of containerd who must have decided to use the TOML format instead of any other.

By hairy here, what I mean is that I don't actually know what to expect in the file from the command. This may be due to my lack of familiarity with what you're describing as a common pattern with sed. The `/a \` in particular -- is that literal, a control sequence, an escape sequence?

I'm not saying that you should do things differently. My command was only intending to ask: "is there a simpler way?".
___
*[manage-cluster/cloud-config_node.yml, line 45 at r2](https://reviewable.io/reviews/m-lab/k8s-support/528#-MOlgobS6gRmSmQZuh-X:-MOlgobS6gRmSmQZuh-Y:b1v8wjc) ([raw file](https://github.com/m-lab/k8s-support/blob/4350049bd29dd9748771721a8154bdb3e1b321e3/manage-cluster/cloud-config_node.yml#L45)):*
> ```YAML
>     net.bridge.bridge-nf-call-ip6tables = 1
> 
> runcmd:
> ```

Should these nodes also `mask` docker?
___
*[manage-cluster/kubeadm-config.yml.template, line 31 at r1](https://reviewable.io/reviews/m-lab/k8s-support/528#-MOhooyV7MNqy8E-T6MA:-MOlfCW01rqjqgF1HZGl:bsa6ap6) ([raw file](https://github.com/m-lab/k8s-support/blob/49033e6c2e2b9d306c21fbde2cd96b0c291886ae/manage-cluster/kubeadm-config.yml.template#L31)):*

I added this `resolv-conf` config here because I happened to observe in the kubelet logs a message stating that the flag `--resolv-conf` was deprecated and that the configuration should be moved to kubelet config file, which this does. I got the value by observing the value of the flag on a running kubelet (kubeadm sets the value).

I actually never knew what that `fix-resolv-conf` initContainer was for.  Someone else (@pboothe ?) created it, and I never really questioned it. Why do we need to much with that.  We can have this conversation elsewhere. I'm just mentioning here that I never knew why that was necessary.
</blockquote></details>

iirc, I added that during the initial new platform demos for the 10th anniversary (when we just had two nodes) to prove we had anything running at all. iirc, this was needed for the containers to perform their own DNS resolution.. and it evidently wasn't working without it. This may have been something as simple as any out bound network connection, such as done by pusher to connect to the storage API, but I think there was an even more basic case that I'm forgetting.

nkinkade (commented):
> *[Reviewable](https://reviewable.io/reviews/m-lab/k8s-support/528)* status: :shipit: complete! 1 of 1 LGTMs obtained

___
*[manage-cluster/bootstraplib.sh, line 210 at r1](https://reviewable.io/reviews/m-lab/k8s-support/528#-MOhpPDM1FxQDPIrZisk:-MOlhduYC-6KVnIQzDTl:b-76j8m2) ([raw file](https://github.com/m-lab/k8s-support/blob/49033e6c2e2b9d306c21fbde2cd96b0c291886ae/manage-cluster/bootstraplib.sh#L210)):*

My "shakes fist" was mostly a joke. The part that wasn't a joke was directed at the maintainers of containerd who must have decided to use the TOML format instead of any other.

By hairy here, what I mean is that I don't actually know what to expect in the file from the command. This may be due to my lack of familiarity with what you're describing as a common pattern with sed. The `/a \` in particular -- is that literal, a control sequence, an escape sequence?

I'm not saying that you should do things differently. My command was only intending to ask: "is there a simpler way?".
</blockquote></details>

I don't know of a simpler way at the moment. If you have some ideas, let me know and I'm very open to simplifying this!

The `sed` pattern I'm using is `'/<match-string>/<action> <text>'`. The `a` <action> means append. The backslash after the `a` is merely to escape a single space, causing `sed` to observe the leading whitespace (which it would otherwise disregard), which I want to maintain the indention in the file. 
___
*[manage-cluster/cloud-config_node.yml, line 45 at r2](https://reviewable.io/reviews/m-lab/k8s-support/528#-MOlgobS6gRmSmQZuh-X:-MOljLzE6Qo5TMCHLAiD:bnd0ves) ([raw file](https://github.com/m-lab/k8s-support/blob/4350049bd29dd9748771721a8154bdb3e1b321e3/manage-cluster/cloud-config_node.yml#L45)):*

Should these nodes also `mask` docker?
</blockquote></details>

Docker doesn't need to be masked because Docker is not even installed!
___
*[manage-cluster/kubeadm-config.yml.template, line 31 at r1](https://reviewable.io/reviews/m-lab/k8s-support/528#-MOhooyV7MNqy8E-T6MA:-MOljYAM5d7PKwtd2xvu:bxeoxc) ([raw file](https://github.com/m-lab/k8s-support/blob/49033e6c2e2b9d306c21fbde2cd96b0c291886ae/manage-cluster/kubeadm-config.yml.template#L31)):*

iirc, I added that during the initial new platform demos for the 10th anniversary (when we just had two nodes) to prove we had anything running at all. iirc, this was needed for the containers to perform their own DNS resolution.. and it evidently wasn't working without it. This may have been something as simple as any out bound network connection, such as done by pusher to connect to the storage API, but I think there was an even more basic case that I'm forgetting.
</blockquote></details>

Maybe we should review that initContainer to find out if it is even needed any longer. I can do some experimenting in sandbox today. And we can chat briefly about it the Open VC.

pboothe:
FYI: https://github.com/google/jsonnet/pull/866 means that modern versions of jsonnet might be able to emit TOML from a jsonnet config.

PR #533, opened 2021-01-08
Updates kured image + adds back --lock-ttl flag.

nkinkade:
Previously, we were using a flag named `--annotation-ttl` to set a timeout for a node to reboot. If the node didn't reboot within that time and remove the lock, then another node could forcefully remove and take the lock so that rolling reboots didn't get hung indefinitely because of one node failing. However, [there was a bug with that flag](https://github.com/weaveworks/kured/issues/143), and we removed it. The bug has since been resolved, apparently, and the flag renamed to `--lock-ttl`. This PR reintroduces that concept using the new flag, and updates kured to the latest version.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/k8s-support/533)

### DIFF
--- a/k8s/daemonsets/core/kured.jsonnet
+++ b/k8s/daemonsets/core/kured.jsonnet
@@ -27,9 +27,7 @@
             args: [
               '--reboot-sentinel=/var/run/mlab-reboot',
               '--period=1h',
-              // Annotation TTL should stay disabled until this bug is close:
-              // https://github.com/weaveworks/kured/issues/143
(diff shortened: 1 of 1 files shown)


stephen-soltesz (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/m-lab/k8s-support/533#-:-MQZC0G8EC5O-cMAduvz:bnfp4nl)* status: :shipit: complete! 1 of 1 LGTMs obtained

PR #538, opened 2021-01-15
Migrates to non-legacy LB backend type + fixes containerd config

### LINKED ISSUE #535: API nodes sometimes (always?) fail to rejoin the API cluster
Just recently I have noticed in sandbox and staging that the first API server in each cluster, after being rebooted, fails to rejoin the API cluster. The error from the kubelet is that it is getting a connection refused from the API load-balancer URL. Wondering if this could be some sort of a hairpin issue in the loadbalancer, I removed the `us-central1-a` (the one failing to rejoin) from the load balancer, and almost immediately it rejoined.

Master nodes used to reboot just fine. It seems that this new behavior may be related in some way to the conversion from Docker to containerd as the CRI for kubernetes.

What we have is a situation where backends of the load-balancer are attempting to communicate with the load-balancer itself, a sort of self-referential problem. It would that the when a backend attempts to communicate with the load-balancer of which it is a part, that the traffic is sent back to that same backend/node. This doesn't work because the kubelet needs to communicate with the API cluster _and_ is also responsible for starting up the api-server pod.

We are currently using a "legacy" TCP/UDP target-pool-based external load balancer. Among other things, this type of backend only supports HTTP health checks, which necessitates the use of the `gcp-loadbalancer-proxy`. Perhaps these issues might go away if we move to the more modern "backend service" type of external load-balancer? It seems worth the trouble, if for nothing else than to get rid of the `gcp-loadbalancer-proxy` and let GCP health checks talk directly to the nodes in the backend instance group using HTTPS.

nkinkade:
While investigating https://github.com/m-lab/k8s-support/issues/535, I realized that we were using a legacy load balancer backend type ("target pool"). In a first attempt to resolve #535 I tried migrating our LB to use a "backend service" backend type instead. It didn't fix the issue, but it had one good result, in that the new backend-service type supports all sorts of health checks (target pools only support "legacy" HTTP checks). This means that we no longer need the [gcp-loadbalancer-proxy](https://github.com/m-lab/gcp-loadbalancer-proxy) container for health checks. The vast majority of this PR is migrating from the legacy "target-pool" backed to the "backend-service" backend.

Additionally, I discovered that in TOML indentation doesn't matter, hence the way I was configuring containerd was incorrect. This PR moves away from using a somewhat cryptic `sed` statement to actually writing the entire config file out using the cloud-init config.

This PR should resolve #535, too.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/k8s-support/538)

### DIFF
--- a/manage-cluster/README.md
+++ b/manage-cluster/README.md
@@ -33,7 +33,7 @@ These steps must be done manually before running the cluster bootstrap script.
 $ ./bootstrap_k8s_master_cluster.sh <gcp-project-name>
 ```
 
-# The ./bootstrap\_k8s\_master\_cluster.sh script
+# The ./bootstrap\_platform\_cluster.sh script
 This is a ridiculously long bash script, but it is not complicated; there are
 just a lot of steps to take and commands to be run. Additionally, line wrapping
 for readability makes is a good deal longer than it might otherwise be.  The
@@ -53,7 +53,7 @@ basic flow of the script boils down to this:
    target-pool. In essence, this forwarding-rule is the load balancer.
 5. Create a firewall rule allowing access to sshd and the k8s-api-server from
    anywhere.
- 
+
 ## Configure internal load balancing
 1. Determine/create an internal IP for the internal load balancer, and update
    Google Cloud DNS accordingly.
@@ -74,21 +74,3 @@ basic flow of the script boils down to this:
 5. Login to the GCE instance and install any necessary packages and
    configurations.
 6. Configure k8s and etcd using `kubeadm`.
-
-# gcp-loadbalancer-proxy
-Our k8s api servers only run over HTTPS, but currently the TCP network load
-balancer in GCP only supports HTTP health checks. Therefore we must run a
-container that will expose an HTTP port and will run an HTTPS request on the
-api-server on the localhost. From
-https://cloud.google.com/load-balancing/docs/network/:
-
-> Network Load Balancing relies on legacy HTTP Health checks for determining
-> instance health. Even if your service does not use HTTP, you'll need to at least
-> run a basic web server on each instance that the health check system can
-> query.
-
-https://github.com/kubernetes/kubernetes/issues/43784#issuecomment-415090237
-
-For this purpose a special purpose
-[gcp-loadbalancer-proxy](https://github.com/m-lab/gcp-loadbalancer-proxy) was
-created. It gets installed automatically in bootstrap\_k8s\_master\_cluster.sh.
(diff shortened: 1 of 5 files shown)


stephen-soltesz (approved):
:lgtm: - small comment.

> Reviewed 5 of 5 files at r1.
> *[Reviewable](https://reviewable.io/reviews/m-lab/k8s-support/538#-:-MR6--RCErHcvXyzDfC8:b-6nixli)* status: :shipit: complete! 1 of 1 LGTMs obtained

___
*[manage-cluster/cloud-config_master.yml, line 189 at r1](https://reviewable.io/reviews/m-lab/k8s-support/538#-MR5zYjjAfXAtl-LDlsY:-MR5zYjjAfXAtl-LDlsZ:bg441xe) ([raw file](https://github.com/m-lab/k8s-support/blob/85e626b9ba6a570daeb94e064f900fb59f5a99a9/manage-cluster/cloud-config_master.yml#L189)):*
> ```YAML
> - systemctl enable docker
> - systemctl start docker
> - systemctl enable gcp-loadbalancer-proxy.service
> ```

When I originally created this, I hoped we would be able to retire it one day. So, I am happy to see it go now.
___
*[manage-cluster/README.md, line 33 at r1](https://reviewable.io/reviews/m-lab/k8s-support/538#-MR5zn4SBorWu2gwdpcH:-MR5zn4SBorWu2gwdpcI:bw4mywd) ([raw file](https://github.com/m-lab/k8s-support/blob/e3e1a67f673aa7404ed430e92c22c4b5e5991453/manage-cluster/README.md#L33)):*
> ```Markdown
> 
> ```bash
> $ ./bootstrap_k8s_master_cluster.sh <gcp-project-name>
> ```

There are additional uses of the old script name in this file. Please update them all. (this may be the only other one :) )

nkinkade (commented):
> *[Reviewable](https://reviewable.io/reviews/m-lab/k8s-support/538)* status: :shipit: complete! 1 of 1 LGTMs obtained

___
*[manage-cluster/README.md, line 33 at r1](https://reviewable.io/reviews/m-lab/k8s-support/538#-MR5zn4SBorWu2gwdpcH:-MR61JqU8hsXjuUEL-YF:b-reaf1c) ([raw file](https://github.com/m-lab/k8s-support/blob/e3e1a67f673aa7404ed430e92c22c4b5e5991453/manage-cluster/README.md#L33)):*

There are additional uses of the old script name in this file. Please update them all. (this may be the only other one :) )
</blockquote></details>

Done. I also read through the whole document and made a few other fixes. I also remove any use of the word "master" in favor of "API".

PR #541, opened 2021-01-21
Use explicit dns config

stephen-soltesz:
This change removes the initContainer that manually overwrites the content of /etc/resolv.conf in favor of using the native k8s support for specifying the `dnsPolicy` and `dnsConfig` -- see also: https://kubernetes.io/docs/concepts/services-networking/dns-pod-service/

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/k8s-support/541)

### DIFF
--- a/k8s/daemonsets/templates.jsonnet
+++ b/k8s/daemonsets/templates.jsonnet
@@ -480,10 +480,12 @@ local Experiment(name, index, bucket, anonMode, datatypes=[]) = ExperimentNoInde
         },
       },
       spec+: {
-        //dnsPolicy: 'None',
-        //dnsConfig: {
-        //  nameservers: ['8.8.8.8'],
-        //},
+        // default kube-dns configuration because M-Lab pod networks bypass
+        // kubernetes Services iptables rules.
+        dnsPolicy: 'None',
+        dnsConfig: {
+          nameservers: ['8.8.8.8'],
+        },
         // Only enable extended grace period where production traffic is possible.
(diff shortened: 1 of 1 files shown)


stephen-soltesz:
See also: https://github.com/m-lab/k8s-support/issues/542

nkinkade (approved):
Do we really want to explicitly and completely override all cluster-local resolution

> *[Reviewable](https://reviewable.io/reviews/m-lab/k8s-support/541#-:-MRar7477jgX_JCbDG2m:b-lp58s3)* status: 0 of 1 LGTMs obtained

*[a discussion](https://reviewable.io/reviews/m-lab/k8s-support/541#-MRfixVdEa1_GkNRm-uP:-MRfixVe4qaK-GvHz_EX:b-xjcbse) (no related file):*
Thank you for this. This is far cleaner and explicit than the initContainer.

---

stephen-soltesz (commented):
I think the answer to your question is: we do not want to override all cluster-local resolution. And, we do not know how to work with the cluster-local resolution yet.

I would prefer (almost always) to work with the system's way of doing things rather than go against the grain and build custom strategies.

In this case, I think we don't know how to do that yet while preserving our multus network configuration for experiments. If we could do both, that would be welcome.

> *[Reviewable](https://reviewable.io/reviews/m-lab/k8s-support/541#-:-MRfr2-i0JpZFtGH3A56:b-fp9sy7)* status: 0 of 1 LGTMs obtained

PR #556, opened 2021-02-12
Updates kube-state-metrics + increase Prometheus VM size

nkinkade:
This PR updates kube-state-metrics and also removes the `addon-resizer` container from the pod. Current documentation for kube-state-metrics makes no mention of this sidecar container and today it was getting oom-killed. Now, the kube-state-metrics container defines its own static resource limits based on [some (possibly outdated) recommendations](https://github.com/kubernetes/kube-state-metrics#resource-recommendation). The current settings consider a node count of roughly 500, and then adds in plenty of breathing room.

We are also generally seeing an increase in the memory usage of Prometheus, and the VM seems to now however at around 80% of it's memory capacity, with spikes near the limit. This PR changes the VM type for Prometheus from n1-highmem-16 to [n2-highmem-16](https://cloud.google.com/compute/docs/machine-types#n2_machine_types), which not only brings in ~25GB more memory but also sports faster CPUs.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/k8s-support/556)

### DIFF
--- a/k8s/deployments/kube-state-metrics.jsonnet
+++ b/k8s/deployments/kube-state-metrics.jsonnet
@@ -30,7 +30,7 @@
             args: [
               '--collectors=daemonsets,deployments,nodes,pods,resourcequotas,services',
             ],
-            image: 'quay.io/coreos/kube-state-metrics:v1.8.0',
+            image: 'k8s.gcr.io/kube-state-metrics/kube-state-metrics:v1.9.7',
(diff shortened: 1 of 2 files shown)


stephen-soltesz (approved):
> *[Reviewable](https://reviewable.io/reviews/m-lab/k8s-support/556)* status: 0 of 1 LGTMs obtained

PR #557, opened 2021-02-17
Add -basicauth subdomain for the platform-cluster Prometheus

robertodauria:
This PR creates the `prometheus-platform-cluster-basicauth` subdomain, protected via HTTP basic authentication and pointing to the same Prometheus service as `prometheus-platform-cluster`.

To work around the 64-character limit LetsEncrypt enforces for the CN field, the two domains are under the same TLS certificate. Only the first one (< 64 characters) in the list is used as CN, while both are listed as "alternative names".  

This is the first step in migrating prometheus-platform-cluster to OAuth. Following PRs will change the prometheus-federation to scrape the -basicauth domain and only after that I'll enable OAuth on the original subdomain, to make sure the system in a working state at all times.

Note: differently from what happens with ingresses created on GKE, the DNS entries must already exist on Google Cloud DNS before deploying these changes as they won't be created for us.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/k8s-support/557)

### DIFF
--- a/k8s/clusterissuers/letsencrypt-staging.jsonnet
+++ b/k8s/clusterissuers/letsencrypt-staging.jsonnet
@@ -26,6 +26,7 @@
           selector: {
             dnsNames: [
               'prometheus-platform-cluster.' + std.extVar('PROJECT_ID') + '.measurementlab.net',
+              'prometheus-platform-cluster-basicauth.' + std.extVar('PROJECT_ID') + '.measurementlab.net',
             ],
           },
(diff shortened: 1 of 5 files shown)


nkinkade (approved):
> Reviewed 5 of 5 files at r1.
> *[Reviewable](https://reviewable.io/reviews/m-lab/k8s-support/557)* status: 0 of 1 LGTMs obtained

PR #558, opened 2021-02-18
Enable OAuth authentication on the platform-cluster-prometheus URL

robertodauria:
Step 3/3: This PR enables OAuth on https://platform-cluster-prometheus.${PROJECT}.measurementlab.net .

Instead of running a separate oauth2-proxy instance, this leverages the existing one running on GKE. Since the session cookie is valid for every subdomain of `${PROJECT}.measurementlab.net` this works without any additional change, and the user does not need to re-login when moving from the main Prometheus instance to the platform-cluster one.

You can test it at https://prometheus-platform-cluster.mlab-sandbox.measurementlab.net

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/k8s-support/558)

### DIFF
--- a/k8s/services/prometheus-tls-ingress.jsonnet
+++ b/k8s/services/prometheus-tls-ingress.jsonnet
@@ -7,9 +7,14 @@
     annotations: {
       'kubernetes.io/tls-acme': 'true',
       'kubernetes.io/ingress.class': 'nginx',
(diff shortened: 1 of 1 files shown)


stephen-soltesz (approved):
> *[Reviewable](https://reviewable.io/reviews/m-lab/k8s-support/558)* status: 0 of 1 LGTMs obtained

PR #567, opened 2021-04-05
Revert traceroute3 datatype change

stephen-soltesz:
This change reverts the datatype directory name change from "traceroute3" back to "traceroute". While the traceroute binary version is conditional on project, these datatype names are not yet. And, we are no yet ready to name traceroute data "traceroute3". As well, with the redesign of the traceroute-caller & hop annotation, the output format may properly remain the same as the current output format (though there may be an additional output datatype).

FYI: @SaiedKazemi

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/k8s-support/567)


nkinkade (approved):
> *[Reviewable](https://reviewable.io/reviews/m-lab/k8s-support/567)* status: 0 of 1 approvals obtained

PR #550, opened 2021-02-02
Installs conntrack, needby by k8s now

nkinkade:
This, too, [was already done in epoxy-images](https://github.com/m-lab/epoxy-images/pull/198/files), and now needs to be done for API nodes, too.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/k8s-support/550)

PR #548, opened 2021-02-02
Updates URLs for kubelet.service and 10-kubeadm.conf

nkinkade:
This change has [already been made in epoxy-images](https://github.com/m-lab/epoxy-images/pull/198/files), and we now need to make it in this repo too. These files must have been _very_ recently removed from the `kubernetes/kubernetes` repo, since deployments worked fine for mlab-sandbox and mlab-staging a week or so ago, but are now failing in production without these changes.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/k8s-support/548)

PR #572, opened 2021-06-08
Fix script to add a new virtual k8s node.

robertodauria:
This PR updates the URL to download `kubelet.service` and `10-kubeadm.conf` since it's been changed upstream and installs conntrack to prevent a warning when running kubeadm.

This change is [Reviewable](https://reviewable.io/reviews/m-lab/k8s-support/572)


nkinkade (approved):
@robertodauria: I actually fixed both of these issues in a separate script earlier this year:

https://github.com/m-lab/k8s-support/pull/550
https://github.com/m-lab/k8s-support/pull/548

For some reason that I can't recall now, I chose to use a static version number for the systemd unit files. I remember having a small conversation with Stephen about it, but can't remember my logic. If I had to guess I'd say it was that the v0.7.0 had the versions of the files we wanted and expected, and since they wouldn't change with a static version, we were safe, and it could be updated when needed.

That said, I think it's fine, and probably better, to have a configurable version string in our configuration file. Three things:

* Do you think it makes sense to rename the variable to something more like `K8S_TOOLING_VERSION`? When I see "RELEASE", I think it has something to do with the current kubernetes release (e.g., v1.21.x).
* Any reason you are using `v0.4.0` of kubernetes/release? In my static versioning from earlier this year, I used `v0.7.0`, but I see that the current release is actually `v0.9.0`.
* Would you also update `manage-cluster/bootstraplib.sh` to include the new variable instead of the static version string?

With the above comments taken into consideration, :lgtm:.

> *[Reviewable](https://reviewable.io/reviews/m-lab/k8s-support/572#-:-MbgMXmk5o1tSDKlqbOm:b-r9pjdw)* status: 0 of 1 approvals obtained

nkinkade:
@robertodauria: I just noticed this old which was approved but never merged. Was there a reason it wasn't merged, or just an oversight?

robertodauria (commented):
> Do you think it makes sense to rename the variable to something more like K8S_TOOLING_VERSION? When I see "RELEASE", I think it has something to do with the current kubernetes release (e.g., v1.21.x).

I can see how it's confusing. I had originally named it after the repo (github.com/kubernetes/release) but it makes more sense to distinguish the tooling from the Kubernetes version.

> Any reason you are using v0.4.0 of kubernetes/release? In my static versioning from earlier this year, I used v0.7.0, but I see that the current release is actually v0.9.0.

That was likely a mistake, or I wanted to take the first version after they migrated those files out of the main Kubernetes repository to minimize the chance that there were differences.
I've checked for changes between v0.4.0, v0.7.0 and the current latest (v0.11.0), those two files are completely unchanged so I'm going with 0.11.0.

> Would you also update manage-cluster/bootstraplib.sh to include the new variable instead of the static version string?

Done. PTAL again?

> *[Reviewable](https://reviewable.io/reviews/m-lab/k8s-support/572#-:-Mjj3Wyn-zy-KxHmKi7F:b2f3h1f)* status: 0 of 1 approvals obtained

nkinkade (approved):
> Reviewed 3 of 3 files at r3.
> *[Reviewable](https://reviewable.io/reviews/m-lab/k8s-support/572)* status: 0 of 1 approvals obtained

PR #577, opened 2021-06-14
Adds ndt-canary DaemonSet to test new ndt-server versions

robertodauria:
This PR creates a new DaemonSet named ndt-canary which is meant to be used to test out new ndt-server releases on a small subset of machines before making it available everywhere.

This DS targets machines with `mlab/ndt-version=canary`. This is not meant to be a definitive solution to canary releases but a stopgap to be able to release ndt-server v0.20.6.

To designate a node as "canary", just run `kubectl label node <node> mlab/ndt-version=canary --overwrite`  on the correct platform cluster.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/k8s-support/577)


stephen-soltesz (changes requested):
> *[Reviewable](https://reviewable.io/reviews/m-lab/k8s-support/577)* status: 0 of 1 approvals obtained (waiting on @robertodauria)

___
*[system.jsonnet, line 22 at r1](https://reviewable.io/reviews/m-lab/k8s-support/577#-McA25M001lN_ppYwia-:-McA25M1CGLyO3JAc2Be:bvem2h4) ([raw file](https://github.com/m-lab/k8s-support/blob/7f264a15811c4d01a091f896a961da9ae428da09/system.jsonnet#L22)):*
> ```Jsonnet
>     import 'k8s/daemonsets/core/node-exporter.jsonnet',
>     import 'k8s/daemonsets/experiments/ndt.jsonnet',
>     import 'k8s/daemonsets/experiments/ndtcloud.jsonnet',
> ```

aside: we should probably name "ndtcloud" "ndt-cloud" instead at some point.
___
*[k8s/daemonsets/experiments/ndt-canary.jsonnet, line 26 at r1](https://reviewable.io/reviews/m-lab/k8s-support/577#-McA33TD31eJ_VCKZinz:-McA33TD31eJ_VCKZio-:bpy4gzv) ([raw file](https://github.com/m-lab/k8s-support/blob/7f264a15811c4d01a091f896a961da9ae428da09/k8s/daemonsets/experiments/ndt-canary.jsonnet#L26)):*
> ```Jsonnet
>       spec+: {
>         nodeSelector: {
>           'mlab/type': 'canary',
> ```

Do I understand correctly, that all other experiments (neubot, wehe, revtr) will not be able to run on "mlab/type:canary" nodes?

If so, does everything else around them work as intended (e.g. monitoring takes these nodes out of rotation for Locate requests)?
___
*[k8s/daemonsets/experiments/ndt-canary.jsonnet, line 42 at r1](https://reviewable.io/reviews/m-lab/k8s-support/577#-McA3akODYmzLP1Lacbn:-McA3akPDOTUS4Mw6jd9:b-86gopw) ([raw file](https://github.com/m-lab/k8s-support/blob/7f264a15811c4d01a091f896a961da9ae428da09/k8s/daemonsets/experiments/ndt-canary.jsonnet#L42)):*
> ```Jsonnet
>               "--",
>             ],
>             args: [
> ```

How do we know that the data from this server is part of the canary pool?

robertodauria:
For some reason Reviewable doesn't allow me to post replies via the Publish button, so I'm doing it the old-fashioned way. :) 


> 
> Do I understand correctly, that all other experiments (neubot, wehe, revtr) will not be able to run on "mlab/type:canary" nodes?
> 
> If so, does everything else around them work as intended (e.g. monitoring takes these nodes out of rotation for Locate requests)?
> 

I've changed this part as discussed.

Now each DS (`production` and `canary`) references a different value of `mlab/ndt-version` explicitly. A separate PR sets `mlab/ndt-version` to `production` by default when a node boots. We are not using `mlab/type` anymore so we don't conflate the machine type with the ndt-server release type. This approach also allows for canary releases for other experiments in a rather straightforward way.

This also changes the answer to your question: other experiments will keep running on canary ndt nodes.

> How do we know that the data from this server is part of the canary pool?

As a short-term solution for this canary release, it will be the only data coming from ndt-server `v0.20.6`. ndt-server version and commit hash are stored in BigQuery.

Summarizing a bit of the long-term plan here:

- We should add a boolean command-line flag like `-canary=true`
- The value of the -canary flag should end up in the data files stored by ndt-server, and consequently be pushed to GCS, parsed and inserted into BigQuery
- To avoid changing the parser, we can leverage the client metadata (which is a free-form key/value map.) and store the `canary` flag there, prefixed in such a way that clients won't override it, OR we could add it as a suffix in the field holding the ndt-server version (e.g. `v0.20.6-canary`)

stephen-soltesz (approved):
Have you manually added the `mlab/ndt-version production` to all the nodes in staging (and prod?) (and sandbox? -- I didn't look, presumably that was necessary to get this to work.)

> Reviewed 1 of 3 files at r1, 2 of 3 files at r2.
> *[Reviewable](https://reviewable.io/reviews/m-lab/k8s-support/577#-:-McF9iqY1Q6QbxNrCNg9:b5xv6gt)* status: 0 of 1 approvals obtained (waiting on @robertodauria)

___
*[k8s/daemonsets/experiments/ndt-canary.jsonnet, line 26 at r1](https://reviewable.io/reviews/m-lab/k8s-support/577#-McA33TD31eJ_VCKZinz:-McF96gj7LNzcOuyOT54:bqvg6zo) ([raw file](https://github.com/m-lab/k8s-support/blob/7f264a15811c4d01a091f896a961da9ae428da09/k8s/daemonsets/experiments/ndt-canary.jsonnet#L26)):*

Do I understand correctly, that all other experiments (neubot, wehe, revtr) will not be able to run on "mlab/type:canary" nodes?

If so, does everything else around them work as intended (e.g. monitoring takes these nodes out of rotation for Locate requests)?
</blockquote></details>

Addressed by new strategy.
___
*[k8s/daemonsets/experiments/ndt-canary.jsonnet, line 42 at r1](https://reviewable.io/reviews/m-lab/k8s-support/577#-McA3akODYmzLP1Lacbn:-McF9ALnAFwFU-QKRCKK:b7mfvja) ([raw file](https://github.com/m-lab/k8s-support/blob/7f264a15811c4d01a091f896a961da9ae428da09/k8s/daemonsets/experiments/ndt-canary.jsonnet#L42)):*

How do we know that the data from this server is part of the canary pool?
</blockquote></details>

Great, your summary is very helpful. The two candidate ideas both use a new flag to ndt-server.

* add reliable information to client metadata (i.e. canary = true)
* add a suffix to the machine version (i.e. "-canary")

For this initial version, we can use the raw version and time.

Is there an issue or bullet in a milestone for adding this feature?

PR #579, opened 2021-06-15
change to scamper (not daemon) and long trace timeout and cache timeout

gfr10598:
It looks like the update to v0.8.0 with 15 minute timeout still has more than 50% timeout running in staging.

This changes the config to use 30 minute timeout, and use scamper instead of scamper-daemon.

This will likely result in lots of zombie processes, but have some idea how to fix that.  Should also follow up with scamper authors to find out their recommended best practices for this kind of application.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/k8s-support/579)

### DIFF
--- a/k8s/daemonsets/templates.jsonnet
+++ b/k8s/daemonsets/templates.jsonnet
@@ -196,7 +196,10 @@ local Traceroute(expName, tcpPort, hostNetwork) = [
       '-uuid-prefix-file=' + uuid.prefixfile,
       '-poll=false',
       '-tcpinfo.eventsocket=' + tcpinfoServiceVolume.socketFilename,
-      '-tracetool=scamper-daemon',
+      '-tracetool=scamper',
+      '-scamper.timeout=30m',
+      '-IPCacheTimeout=30m',
+      '-IPCacheUpdatePeriod=1m'
     ],
     env: if hostNetwork then [] else [
       {


SaiedKazemi:
As you noticed, running "scamper" instead of "scamper-daemon" causes defunct processes.  On my 64-bit machine, /proc/sys/kernel/pid_max is 4,194,304 which, depending on traffic, can be consumed quickly.

How about we fix the defunct issue first and then switch to scamper?

stephen-soltesz (commented):
I concur with @saiedkazemi  -- I would like a better understanding of why we preferred scamper-daemon originally (I imagine it is preferable for forgotten reasons) -- and there may be options that address the timeouts you're observing.

If scamper binary is still the way to go, the zombie process leak should be addressable using a strategy with `tini` (a mini init-like substitute for docker images -- it's main job is to reap disowned children).

See: https://github.com/m-lab/script-exporter-support/blob/master/Dockerfile.minimal#L16

> *[Reviewable](https://reviewable.io/reviews/m-lab/k8s-support/579#-:-McFUZnvCHjgiJM-uaBm:b7dzqk1)* status: 0 of 1 approvals obtained (waiting on @SaiedKazemi)

stephen-soltesz (commented):
@gfr10598 to frame this differently -- there is a notion of "operational health" that's analogous to "code health" that could apply here.

> *[Reviewable](https://reviewable.io/reviews/m-lab/k8s-support/579#-:-McF_CJ83gIihB5ZPU_8:b2x7l0g)* status: 0 of 1 approvals obtained (waiting on @SaiedKazemi)

nkinkade:
@gfr10598 (and maybe @SaiedKazemi): has this PR been superseded by other work done to TRC? If this PR is no longer valid, then would someone please close it?

SaiedKazemi:
@gfr10598 I am closing this PR because TRC is already running an instance of the scamper binary for each traceroute (i.e., doesn't use scamper in daemon mode anymore).

PR #602, opened 2021-08-12
Sandbox kinkade

nkinkade:
Not every DaemonSet pod is going to have a "net1" interface, since not all of them use ipvlan. In those cases, the current initContainer will exit with a non 0 status, and then crash loop. This PR instead uses `sh` to just splat a 0 value to all existing interfaces for `/proc/sys/net/ipv6/conf/*/{accept_ra,autoconf}`.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/k8s-support/602)

### DIFF
--- a/k8s/daemonsets/templates.jsonnet
+++ b/k8s/daemonsets/templates.jsonnet
@@ -52,7 +52,7 @@ local uuid = {
 local sysctls = {
   initContainer: {
     command: [
-      'sysctl', '--write', '--ignore',
+      'sysctl', '-e', '-w',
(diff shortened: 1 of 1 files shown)


stephen-soltesz (approved):
:lgtm: -- thank you for the fix.

> *[Reviewable](https://reviewable.io/reviews/m-lab/k8s-support/602#-:-MgvUJ5s0_jPobpqVeeK:b-pfwbsp)* status: 0 of 1 approvals obtained

PR #601, opened 2021-08-11
Adds privileged initContainer for setting sysctls

nkinkade:
Kubernetes supports a mechanism for setting an exceedingly small set of
sysctls in a Pod spec:
https://kubernetes.io/docs/tasks/administer-cluster/sysctl-cluster/
... but the sysctls we need to set are not in that list, and the
workaround (using a cluster PodSecurityPolicy) is deprecated and will be
removed from a later version of Kubernetes (v1.25.x):
https://kubernetes.io/blog/2021/04/06/podsecuritypolicy-deprecation-past-present-and-future/

This PR should resolve https://github.com/m-lab/epoxy-images/issues/209

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/k8s-support/601)

PR #603, opened 2021-08-24
Use CNI `tuning` plugin to set various sysctls 

nkinkade:
There is an [open issue](https://github.com/m-lab/epoxy-images/issues/209) to address the case where IPv6 SLAAC and RAs modify default IPv6 routes inside of experiment containers. We _thought_ we had resolved that issue with PR #601, but even after that change was deployed to production, we still discovered random IPv6 breakage. Indeed, you could login to an affected container and show that `net.ipv6.conf.net1.{accept_ra,autoconf}` were both set to 0, yet the default IPv6 route was still gone.

I came across [an old issue](https://github.com/lxc/lxd/issues/3582) that determined that there was a race between SLAAC and network configurations being done from within containers. I believe this is the most plausible explanation of what we are currently seeing. Probably the network namespace is created, and sometime later a process is created using that net-ns (in which our previous initContainer ran), but somewhere between creation of the net-ns and the container SLAAC did its thing and configured a default IPv6 with an expiration.

Investigating this I stumbled across a CNI plugin I had not previously known about  named `tuning`, which allows you to set various network-related kernel parameters. This PR leverages that plugin to set the sysctls required to disabled SLAAC and acceptance of RAs, obviating the need for the initContainer. The hope is that this will set the desired sysctls at an earlier stage, before SLAAC can actually do anything to corrupt our desired configuration.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/k8s-support/603)

### DIFF
--- a/k8s/daemonsets/templates.jsonnet
+++ b/k8s/daemonsets/templates.jsonnet
@@ -46,30 +46,6 @@ local uuid = {
   },
 };
 
-// An initContainer that can be used to set various sysctls in pods.
-// NOTE: do not set unnamespaced sysctls here!
-// https://kubernetes.io/docs/tasks/administer-cluster/sysctl-cluster/#setting-sysctls-for-a-pod
-local sysctls = {
-  initContainer: {
-    command: [
-      // Do not use IPv6 autoconfiguration (SLAAC), and reject RAs (Router
-      // Advertisements). When accept_ra=1, RAs can cause the IPv6 network
-      // stack to reconfigure itself, for example changing or removing the
-      // default route.
-      'sh', '-c', |||
-        for i in /proc/sys/net/ipv6/conf/*/accept_ra; do echo 0 > $i; done;
-        for i in /proc/sys/net/ipv6/conf/*/autoconf; do echo 0 > $i; done;
-      |||,
-    ],
-    image: 'busybox',
-    name: 'set-pod-sysctls',
-    securityContext: {
-      privileged: true,
-      runAsUser: 0,
-    },
-  },
-};
-
 local volume(name) = {
   hostPath: {
     path: '/cache/data/' + name,
@@ -468,7 +444,6 @@ local ExperimentNoIndex(name, bucket, anonMode, datatypes, hostNetwork) = {
         [if hostNetwork then 'serviceAccountName']: 'kube-rbac-proxy',
         initContainers: [
           uuid.initContainer,
-          sysctls.initContainer,
         ],
         nodeSelector: {
           'mlab/type': 'physical',
--- a/k8s/networks/networks.jsonnet
+++ b/k8s/networks/networks.jsonnet
@@ -57,18 +57,18 @@
         name: 'ipvlan-index-' + index,
         plugins: [
           {
+            type: 'tuning',
+            sysctl: {
+              'net.ipv6.conf.net1.accept_ra': '0',
+              'net.ipv6.conf.net1.autoconf': '0',
+            },
+          },
+          {
             type: 'ipvlan',
             master: 'eth0',
             ipam: {
               type: 'index2ip',
               index: index,
-            },
-          },
-          {
-            type: 'tuning',
-            sysctl: {
-              'net.ipv6.conf.net1.accept_ra': '0',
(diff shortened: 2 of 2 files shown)


stephen-soltesz (approved):
:lgtm: -- great discovery. one question.

> *[Reviewable](https://reviewable.io/reviews/m-lab/k8s-support/603#-:-MhttHtQ0JdPSCJRPOLm:b-d2osog)* status: 0 of 1 approvals obtained (waiting on @nkinkade)

___
*[k8s/networks/networks.jsonnet, line 67 at r1](https://reviewable.io/reviews/m-lab/k8s-support/603#-Mhtt-tuDepSKfBMn8Ts:-Mhtt-tv1jua5rhV9yDk:b-30qqi5) ([raw file](https://github.com/m-lab/k8s-support/blob/01ad791ef5eeff14a6e22bed9b21594a1e579e6c/k8s/networks/networks.jsonnet#L67)):*
> ```Jsonnet
>             },
>           },
>           {
> ```

Does order matter here? Just wondering out loud. I half expected `tuning` to be first. but maybe it must be run after the ipvlan exists? A short comment would be welcome if we know for sure.

nkinkade (commented):
> *[Reviewable](https://reviewable.io/reviews/m-lab/k8s-support/603)* status: 0 of 1 approvals obtained (waiting on @stephen-soltesz)

___
*[k8s/networks/networks.jsonnet, line 67 at r1](https://reviewable.io/reviews/m-lab/k8s-support/603#-Mhtt-tuDepSKfBMn8Ts:-Mhu0WcoCuIuL96CBVe9:b-376zj3) ([raw file](https://github.com/m-lab/k8s-support/blob/01ad791ef5eeff14a6e22bed9b21594a1e579e6c/k8s/networks/networks.jsonnet#L67)):*
Your question is a good one. I had read that each plugin is run in the order in which it is specified in the config, and that perhaps the output of each plugin is fed as stdin to the next plugin, or at least as some env variable. The few examples I could drum up of using this seemingly lesser known CNI plugin, all seemed to have the `tuning` plugin called after the `ipam` plugin, but that may have just been coincidence.  In the [CNI spec](https://github.com/containernetworking/cni/blob/master/SPEC.md#lifecycle--ordering), I found this:

> The container runtime must create a new network namespace for the container before invoking any plugins.

I gather that as soon as the net-ns exists, that one can start setting the various sysctls for it. In this case, if we can set the sysctls before the ipvlan interface is even created, then we totally eliminate the possibility of a race between SLAAC and anything the `tuning` plugin might do.  I have moved the `tuning` plugin to be first.

stephen-soltesz (commented):
> *[Reviewable](https://reviewable.io/reviews/m-lab/k8s-support/603)* status: 0 of 1 approvals obtained

___
*[k8s/networks/networks.jsonnet, line 67 at r1](https://reviewable.io/reviews/m-lab/k8s-support/603#-Mhtt-tuDepSKfBMn8Ts:-Mhu6bfv5e3b6gCdnNid:bxjpa5s) ([raw file](https://github.com/m-lab/k8s-support/blob/01ad791ef5eeff14a6e22bed9b21594a1e579e6c/k8s/networks/networks.jsonnet#L67)):*

Your question is a good one. I had read that each plugin is run in the order in which it is specified in the config, and that perhaps the output of each plugin is fed as stdin to the next plugin, or at least as some env variable. The few examples I could drum up of using this seemingly lesser known CNI plugin, all seemed to have the `tuning` plugin called after the `ipam` plugin, but that may have just been coincidence.  In the [CNI spec](https://github.com/containernetworking/cni/blob/master/SPEC.md#lifecycle--ordering), I found this:

> The container runtime must create a new network namespace for the container before invoking any plugins.

I gather that as soon as the net-ns exists, that one can start setting the various sysctls for it. In this case, if we can set the sysctls before the ipvlan interface is even created, then we totally eliminate the possibility of a race between SLAAC and anything the `tuning` plugin might do.  I have moved the `tuning` plugin to be first.
</blockquote></details>

Excellent. If this is true, please add a comment to the configuration so we don't lose this insight. :+1:

PR #629, opened 2021-12-15
Sets resolvConf field in KubeletConfiguration

nkinkade:
A couple times now we have seen a CoreDNS pod crash looping with an error about a forwarding loop for the root domain. [The documentation for the 'loop' plugin](https://coredns.io/plugins/loop/#troubleshooting-loops-in-kubernetes-clusters), which detects this, suggests changing the value of resolvConf in the Kubelet configuration from /etc/resolv.conf to
/run/systems/resolv/resolv.conf on systems that are using systemd-resolved, which we do.

This is intended to hopefully resolve issues like these:

https://github.com/m-lab/ops-tracker/issues/1480
https://github.com/m-lab/ops-tracker/issues/1519

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/k8s-support/629)

### DIFF
--- a/manage-cluster/kubeadm-config.yml.template
+++ b/manage-cluster/kubeadm-config.yml.template
@@ -5,7 +5,6 @@ nodeRegistration:
   kubeletExtraArgs:
     container-runtime: "remote"
     container-runtime-endpoint: "unix:///run/containerd/containerd.sock"
-    resolv-conf: "/run/systemd/resolve/resolv.conf"
 localAPIEndpoint:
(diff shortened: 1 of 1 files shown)


stephen-soltesz (approved):
Okay, I'm trusting you on this one. :lgtm:

> *[Reviewable](https://reviewable.io/reviews/m-lab/k8s-support/629#-:-Mr-G3lm3qg-aNN6216N:b-n0j1ue)* status: 0 of 1 approvals obtained

PR #641, opened 2022-01-27
Adds documentation on creating NDT virtual sites

nkinkade:
Also renames the script add_ndt_virtual_node.sh to
add_ndt_virtual_site.sh, which is a bit more accurate.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/k8s-support/641)


stephen-soltesz (approved):
One request, several things to consider for the future.

:lgtm:

> *[Reviewable](https://reviewable.io/reviews/m-lab/k8s-support/641#-:-MuR5Ot84x6tOAAiUMyV:bolqakc)* status: 0 of 1 approvals obtained (waiting on @nkinkade)

___
*[manage-cluster/NDT_VIRTUAL_SITES.md, line 5 at r1](https://reviewable.io/reviews/m-lab/k8s-support/641#-MuQYzWm1_gpQqzlXdEY:-MuQYzWm1_gpQqzlXdEZ:b-mf2zx2) ([raw file](https://github.com/m-lab/k8s-support/blob/df0ac39cdf11c2e043c587b85253465cb588d683/manage-cluster/NDT_VIRTUAL_SITES.md#L5)):*
> ```Markdown
> NDT virtual sites are sites that run only NDT in cloud environments (e.g., GCP, AWS,
> etc.). Differing from their physical counterparts, virtual sites contain only
> a single node (mlab1) with a single IPv4 address (assigned by the cloud provider).
> ```

Just thinking out loud and ahead: we'll probably want more than one node per GCP region at some future point and our naming convention may limit how many "sites" we create in a region. I'm also reminded that at one time (possibly still) we had a hard limit on our current k8s platform cluster config of managing up to 1024 nodes (deriving from a private network size, iirc). This too may not be far off with cloud in the mix.

Also, I'm wondering wether we could target GKE for single-service deployments like ndt (with sidecars). This might make it possible to support auto scaling features in response to demand.

But all of these strategies would require alternate considerations for individual server/site naming.
___
*[manage-cluster/NDT_VIRTUAL_SITES.md, line 10 at r1](https://reviewable.io/reviews/m-lab/k8s-support/641#-MuR3NCe4jVlXopwVQxG:-MuR3NCe4jVlXopwVQxH:b-5yca1h) ([raw file](https://github.com/m-lab/k8s-support/blob/df0ac39cdf11c2e043c587b85253465cb588d683/manage-cluster/NDT_VIRTUAL_SITES.md#L10)):*
> ```Markdown
> Platform (GCE).
> 
> **NOTE**: if you are creating a virtual site in mlab-staging or mlab-oti, and
> ```

Please mention the github maintenance exporter service by name here, if there is a README with documentation about how it operates, linking to that for more information would be welcomed.
___
*[manage-cluster/NDT_VIRTUAL_SITES.md, line 27 at r1](https://reviewable.io/reviews/m-lab/k8s-support/641#-MuR3rov3VZs9b3pyxek:-MuR3rov3VZs9b3pyxel:bp0vv4j) ([raw file](https://github.com/m-lab/k8s-support/blob/df0ac39cdf11c2e043c587b85253465cb588d683/manage-cluster/NDT_VIRTUAL_SITES.md#L27)):*
> ```Markdown
> Unlike physical sites, the machine (VM) for a virtual site must be created in
> the cloud _before_ an entry in [siteinfo](https://github.com/m-lab/siteinfo/) is
> created. This is because we don't know the IP address of the VM prior to its
> ```

It is possible to allocate IPs in GCP independently of the GCE VM creation, and assign them at VM creation time. This may be a way to approach the standard order of events (register, create, deploy, enable).

The IP naming convention could be such that the parameters given to add_ndt_virtual_site.sh are able to automatically discover and select pre-allocated IPs by name.

nkinkade (commented):
> *[Reviewable](https://reviewable.io/reviews/m-lab/k8s-support/641)* status: 0 of 1 approvals obtained (waiting on @stephen-soltesz)

___
*[manage-cluster/NDT_VIRTUAL_SITES.md, line 5 at r1](https://reviewable.io/reviews/m-lab/k8s-support/641#-MuQYzWm1_gpQqzlXdEY:-MuRjHJY7k7Jx52Eq5cc:bz5e1n7) ([raw file](https://github.com/m-lab/k8s-support/blob/df0ac39cdf11c2e043c587b85253465cb588d683/manage-cluster/NDT_VIRTUAL_SITES.md#L5)):*

Just thinking out loud and ahead: we'll probably want more than one node per GCP region at some future point and our naming convention may limit how many "sites" we create in a region. I'm also reminded that at one time (possibly still) we had a hard limit on our current k8s platform cluster config of managing up to 1024 nodes (deriving from a private network size, iirc). This too may not be far off with cloud in the mix.

Also, I'm wondering wether we could target GKE for single-service deployments like ndt (with sidecars). This might make it possible to support auto scaling features in response to demand.

But all of these strategies would require alternate considerations for individual server/site naming.
</blockquote></details>

With out proposed naming structure of virtual nodes having a suffix of "c", then we are limited to 10 virtual nodes in a given GCP region (abc[0-9]c). If we were to decided to not use the "c" suffix idea, then I suppose we could have 100 virtual nodes in a given GCP region (abc[0-9]{2}).

Our current "cluster CIDR" is 192.168.0.0/16, and we allocate a /26 for each node. You are right that this limits us to 1024 nodes, with no more than ~64 pods per node. We have currently used up about 50% of our usable node count. We would have to deploy another 500+ nodes (virtual and physical) to come up against our current limit. Once we start to approach that limit, I imagine we could fairly easily use a different  cluster CIDR, say, 10.0.0.0/8. With something like that we could have, for example, a node CIDR of /24. This would yield us 65,536 possible nodes, each running up to 256 pods. If my memory serves me, the reason we didn't choose to go with a 10.0.0.0 address space to begin with was to avoid any confusion with the default address space GCP uses for VMs, which I believe is also 10.0.0.0 (though this is, I believe, configurable on a per-VPC network).

Running nodes in a completely different kubernetes cluster than the platform cluster is an interesting idea. It might complicate monitoring and alerting, as well as administration, but I imagine could be doable.
___
*[manage-cluster/NDT_VIRTUAL_SITES.md, line 10 at r1](https://reviewable.io/reviews/m-lab/k8s-support/641#-MuR3NCe4jVlXopwVQxG:-MuRpK7Z9xVy-_tXE6TE:b-896fix) ([raw file](https://github.com/m-lab/k8s-support/blob/df0ac39cdf11c2e043c587b85253465cb588d683/manage-cluster/NDT_VIRTUAL_SITES.md#L10)):*

Please mention the github maintenance exporter service by name here, if there is a README with documentation about how it operates, linking to that for more information would be welcomed.
</blockquote></details>

Done.
___
*[manage-cluster/NDT_VIRTUAL_SITES.md, line 27 at r1](https://reviewable.io/reviews/m-lab/k8s-support/641#-MuR3rov3VZs9b3pyxek:-MuRlT6M6yWhO0zBMcAM:b8jcwe8) ([raw file](https://github.com/m-lab/k8s-support/blob/df0ac39cdf11c2e043c587b85253465cb588d683/manage-cluster/NDT_VIRTUAL_SITES.md#L27)):*

It is possible to allocate IPs in GCP independently of the GCE VM creation, and assign them at VM creation time. This may be a way to approach the standard order of events (register, create, deploy, enable).

The IP naming convention could be such that the parameters given to add_ndt_virtual_site.sh are able to automatically discover and select pre-allocated IPs by name.
</blockquote></details>

The script `add_ndt_virtual_site.sh` is just a wrapper around the more generic script `add_k8s_virtual_node.sh`. `add_k8s_virtual_node.sh` already allows you to specify a named GCP IP address.  In this case, `add_ndt_virtual_site.sh` creates the named IP address and passes it to `add_k8s_virtual_node.sh`. 

My first impression, which could be wrong or shortsighted, is that turning the creation of the IP address into separate step just adds another manual step for an operator. The operator would need to pick a region for the IP address, and then remember to use that exact same region for the creation of the virtual site. The script tries to reduce the number of steps necessary for an operator. Currently, there are two: 1) create the virtual site 2) add the site siteinfo.

Is there a benefit that I'm not seeing in moving creation of the IP address to a separate step?

stephen-soltesz (commented):
> *[Reviewable](https://reviewable.io/reviews/m-lab/k8s-support/641)* status: 0 of 1 approvals obtained (waiting on @stephen-soltesz)

___
*[manage-cluster/NDT_VIRTUAL_SITES.md, line 27 at r1](https://reviewable.io/reviews/m-lab/k8s-support/641#-MuR3rov3VZs9b3pyxek:-MuRqqarBtuuLKK68_zH:bmwjt8s) ([raw file](https://github.com/m-lab/k8s-support/blob/df0ac39cdf11c2e043c587b85253465cb588d683/manage-cluster/NDT_VIRTUAL_SITES.md#L27)):*

The script `add_ndt_virtual_site.sh` is just a wrapper around the more generic script `add_k8s_virtual_node.sh`. `add_k8s_virtual_node.sh` already allows you to specify a named GCP IP address.  In this case, `add_ndt_virtual_site.sh` creates the named IP address and passes it to `add_k8s_virtual_node.sh`. 

My first impression, which could be wrong or shortsighted, is that turning the creation of the IP address into separate step just adds another manual step for an operator. The operator would need to pick a region for the IP address, and then remember to use that exact same region for the creation of the virtual site. The script tries to reduce the number of steps necessary for an operator. Currently, there are two: 1) create the virtual site 2) add the site siteinfo.

Is there a benefit that I'm not seeing in moving creation of the IP address to a separate step?
</blockquote></details>

It could allow reserving IPs already registered with siteinfo separately from bringing the VM up (or down). There's a modest charge for that but it could be helpful to quickly increase regional capacity if necessary. I don't want to make things more complicated. :)

nkinkade (commented):
> *[Reviewable](https://reviewable.io/reviews/m-lab/k8s-support/641)* status: 0 of 1 approvals obtained (waiting on @stephen-soltesz)

___
*[manage-cluster/NDT_VIRTUAL_SITES.md, line 27 at r1](https://reviewable.io/reviews/m-lab/k8s-support/641#-MuR3rov3VZs9b3pyxek:-MuRrg3R-Q-GXXTHT6IW:b6wkmwn) ([raw file](https://github.com/m-lab/k8s-support/blob/df0ac39cdf11c2e043c587b85253465cb588d683/manage-cluster/NDT_VIRTUAL_SITES.md#L27)):*

It could allow reserving IPs already registered with siteinfo separately from bringing the VM up (or down). There's a modest charge for that but it could be helpful to quickly increase regional capacity if necessary. I don't want to make things more complicated. :) 
</blockquote></details>

Okay, got it, I think. You are imaging a case where we create "inactive" virtual sites in siteinfo, which we may or may not use one day. But that if one day we decided that we need emergency capacity, we could spin up the VM and everything else is already in place (namely DNS and the site included in the siteinfo API formats). I suppose I get that, but it doesn't really save us much more time the current method, plus an emergency release of siteinfo. In any case, something for consideration.

PR #644, opened 2022-02-07
Configures Vector to export internal metrics

nkinkade:
The changes in this PR cause Vector to export internal metrics about its own operation. We can probably use some of these metrics to determine if Vector is working as intended, and to alert when something seems off.

This is an interesting configuration, as Vector treats its own internal operating metrics in the same way as it does all other data sources and sinks. That is, internal metrics are just another source of data, which can be transformed and shuttled off to some other location. In this case, they aren't shuttled anywhere, but are just served up at `:9090/metrics` inside the pod, which gets picked up automatically by Prometheus kubernetes service discovery.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/k8s-support/644)

### DIFF
--- a/config/vector/values.yaml.template
+++ b/config/vector/values.yaml.template
@@ -59,7 +59,7 @@ extraVolumeMounts:
 image:
   tag: {{IMAGE}}
 podAnnotations:
-  prometheus.io/scrape: true
+  prometheus.io/scrape: "true"
 podLabels:
   workload: vector
 role: Agent
--- a/config/vector/values.yaml.template
+++ b/config/vector/values.yaml.template
@@ -15,11 +15,18 @@ customConfig:
         location: {{PROJECT}}
         namespace: "${VECTOR_SELF_NODE_NAME}"
         node_id: "${VECTOR_SELF_NODE_NAME}"
+    prometheus_exporter:
+      type: prometheus_exporter
+      inputs:
+      - internal_metrics
+      address: 0.0.0.0:9598
   sources:
     journald:
       type: journald
     kubernetes_logs:
       type: kubernetes_logs
+    internal_metrics:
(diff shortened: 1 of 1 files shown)


stephen-soltesz (approved):
> *[Reviewable](https://reviewable.io/reviews/m-lab/k8s-support/644)* status: 0 of 1 approvals obtained

___
*[config/vector/values.yaml.template, line 22 at r1](https://reviewable.io/reviews/m-lab/k8s-support/644#-MvOFzc76ZvYYVTgpJRH:-MvOFzc76ZvYYVTgpJRI:bwmfl0p) ([raw file](https://github.com/m-lab/k8s-support/blob/f67288b27d50d1f77f2922c7597f8a881715c2de/config/vector/values.yaml.template#L22)):*
> ```
>       inputs:
>       - internal_metrics
>       address: 0.0.0.0:9090
> ```

Just to confirm - this is not publicly accessible?

nkinkade (commented):
> *[Reviewable](https://reviewable.io/reviews/m-lab/k8s-support/644)* status: 0 of 1 approvals obtained

___
*[config/vector/values.yaml.template, line 22 at r1](https://reviewable.io/reviews/m-lab/k8s-support/644#-MvOFzc76ZvYYVTgpJRH:-MvPNCaX2VoMHg_Tz6lh:bb9ewp) ([raw file](https://github.com/m-lab/k8s-support/blob/f67288b27d50d1f77f2922c7597f8a881715c2de/config/vector/values.yaml.template#L22)):*

Just to confirm - this is not publicly accessible?
</blockquote></details>

Vector pods do not get a public IP, but only a private cluster IP, so the listener should only be on the loopback and the private cluster IP. I think this should be safe.

PR #645, opened 2022-02-08
Adds speed label to max_by() Prom query

nkinkade:
We use this recording rule to check for switches that are overloaded,
and whether a switch is overloaded depends on how fast the uplink is.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/k8s-support/645)

### DIFF
--- a/config/prometheus/rules.yml
+++ b/config/prometheus/rules.yml
@@ -188,7 +188,7 @@ groups:
   #
   # Units: bits per second.
   - record: switch:ifHCOutOctets:bps2m
-    expr: max by (site, speed) (8 * irate(ifHCOutOctets{ifAlias="uplink"}[4m]))
+    expr: max by (site, speed, ifAlias) (8 * irate(ifHCOutOctets{ifAlias="uplink"}[4m]))
 
   ## NDT Early Warning aggregation rules for Kubernetes platform.
 
--- a/config/prometheus/rules.yml
+++ b/config/prometheus/rules.yml
@@ -188,7 +188,7 @@ groups:
   #
   # Units: bits per second.
   - record: switch:ifHCOutOctets:bps2m
-    expr: max by (site) (8 * irate(ifHCOutOctets{ifAlias="uplink"}[4m]))
+    expr: max by (site, speed) (8 * irate(ifHCOutOctets{ifAlias="uplink"}[4m]))
 
   ## NDT Early Warning aggregation rules for Kubernetes platform.
 


stephen-soltesz (approved):
:lgtm: w/ suggestions

> *[Reviewable](https://reviewable.io/reviews/m-lab/k8s-support/645#-:-MvPXYI9Deul06Xu-Pp_:b4db4wm)* status: 0 of 1 approvals obtained (waiting on @nkinkade)

___
*[config/prometheus/rules.yml, line 191 at r1](https://reviewable.io/reviews/m-lab/k8s-support/645#-MvPXNTbCeLXNKvgNXYo:-MvPXNTbCeLXNKvgNXYp:bwlgsgx) ([raw file](https://github.com/m-lab/k8s-support/blob/c72f42f0306c2a097e1eba3d5d74bc3ed58b94fd/config/prometheus/rules.yml#L191)):*
> ```YAML
>   # Units: bits per second.
>   - record: switch:ifHCOutOctets:bps2m
>     expr: max by (site, speed) (8 * irate(ifHCOutOctets{ifAlias="uplink"}[4m]))
> ```

if we want to keep the SRE dashboard queries unmodified, you can group by `ifAlias` also so that's carried through. Since we're selecting on only "uplink" that's the only one there should be in the result.

nkinkade (commented):
> *[Reviewable](https://reviewable.io/reviews/m-lab/k8s-support/645)* status: 0 of 1 approvals obtained (waiting on @stephen-soltesz)

___
*[config/prometheus/rules.yml, line 191 at r1](https://reviewable.io/reviews/m-lab/k8s-support/645#-MvPXNTbCeLXNKvgNXYo:-MvPZNmW6Ecg-CIu3rSa:b-flbima) ([raw file](https://github.com/m-lab/k8s-support/blob/c72f42f0306c2a097e1eba3d5d74bc3ed58b94fd/config/prometheus/rules.yml#L191)):*

if we want to keep the SRE dashboard queries unmodified, you can group by `ifAlias` also so that's carried through. Since we're selecting on only "uplink" that's the only one there should be in the result.
</blockquote></details>

Good point. I didn't group by `ifAlias`, since we are already selecting on it in the query. My plan was to simply remove the `ifAlias="uplink"` part of the query from the dashboard since it's superfluous. However, I guess there is no harm in grouping on it, and, as you point out, it could mean we wouldn't have to really modify the dashboard, except to change the datasource.

PR #646, opened 2022-02-09
Set -txcontroller.max-rate=150000000 for virtual nodes

nkinkade:
This is normally set at runtime on physical machines based the value in
the ConfigMap nodes-max-rate. However, that ConfigMap is created by
looking at the switch configuration for a site, but virtual sites do not
have switches and thus no entry for switches.json in siteinfo. This
commit manually configures the value to the same value we use for other
1g sites.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/k8s-support/646)


stephen-soltesz (approved):
hmm...

:lgtm:

> *[Reviewable](https://reviewable.io/reviews/m-lab/k8s-support/646#-:-MvUpmbbELzOC9vsUL82:bbxnhpz)* status: 0 of 1 approvals obtained

PR #647, opened 2022-02-09
Removes socat proxy containers

nkinkade:
socat proxy is currently used to access natively instrumented metrics and pprof data for M-Lab core service containers. However, we discovered that socat proxy was not actually needed. Instead we can leverage `kubectl proxy` to make calls directly to the private cluster IPs/ports used by these services. This removes a considerable amount of clutter in our pods, since we will no longer need a separate socat proxy container for every core service container. The documentation for accessing this data without socat proxy was also updated.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/k8s-support/647)

### DIFF
--- a/config/prometheus/prometheus.yml
+++ b/config/prometheus/prometheus.yml
@@ -101,14 +101,6 @@ scrape_configs:
         regex: https
         action: drop
 
-      # socat proxy enables easy access to pprof targets but accidentally
-      # duplicates metric collection. The annotation mechanism applies at the pod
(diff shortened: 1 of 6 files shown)


stephen-soltesz (approved):
:lgtm: -- thank you for this!

> Reviewed 6 of 6 files at r1.
> *[Reviewable](https://reviewable.io/reviews/m-lab/k8s-support/647#-:-MvW4xM067mMwgXQRAZx:bryna78)* status: 0 of 1 approvals obtained

PR #651, opened 2022-02-18
Adds support for adding IPv6-enabled virtual nodes

nkinkade:
Note: dual stack GCE VMs are only available in a few regions. As of this
commit, the only region in the US is us-west2 (LAX). With these changes,
if a VM is being added in one of the few regions where IPv6 is
supported, then this script will automatically enable IPv6.

This PR also updates the default VM machine-type to n1-highcpu-4, which gives these VMs a bit more horsepower.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/k8s-support/651)

### DIFF
--- a/manage-cluster/add_k8s_virtual_node.sh
+++ b/manage-cluster/add_k8s_virtual_node.sh
@@ -21,6 +21,10 @@ usage() {
   echo "USAGE: $0 -p <project> [-z <gcp-zone>] [-m <machine-type>] [-n <gce-name>] [-H <hostname>] [-a <address>] [-t <gce-tag> ...] [-l <k8s-label> ...]"
 }
 
+# List of GCP regions which support external IPv6 addresses for GCE VMs:
+# https://cloud.google.com/vpc/docs/vpc#ipv6-regions
+IPV6_REGIONS="asia-east1 asia-south1 europe-west2 us-west2"
+
 ADDRESS=""
 HOST_NAME=""
 LABELS=""
@@ -115,7 +119,13 @@ else
 fi
 
 if [[ -z "${MACHINE_TYPE}" ]]; then
-  MACHINE_TYPE="n1-standard-1"
+  MACHINE_TYPE="n1-highcpu-4"
+fi
+
+if echo $IPV6_REGIONS | grep ${!GCE_REGION}; then
+  IPV6_FLAG="--stack-type=IPV4_IPV6"
+else
+  IPV6_FLAG=""
 fi
 
 # Allocate a new VM.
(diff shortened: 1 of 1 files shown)


stephen-soltesz (changes requested):
> *[Reviewable](https://reviewable.io/reviews/m-lab/k8s-support/651)* status: 0 of 1 approvals obtained (waiting on @nkinkade)

___
*[manage-cluster/add_k8s_virtual_node.sh, line 299 at r1](https://reviewable.io/reviews/m-lab/k8s-support/651#-MwXZsmXBU2zT5b4YsxZ:-MwXZsmXBU2zT5b4Ysx_:bt8yl35) ([raw file](https://github.com/m-lab/k8s-support/blob/655e336b5f956bda2997ef76b2f65650f4c8ff57/manage-cluster/add_k8s_virtual_node.sh#L299)):*
> ```Shell
>   echo -n $EXTERNAL_IP > "\${metadata_dir}/external-ip"
>   if [[ -n $EXTERNAL_IPV6 ]]; then
>     echo -n $EXTERNAL_IPV6 > "\${metadata_dir}/external-ipv6"
> ```

Google get's a C for consistently naming things.

We will need a way to plumb this value into an additional label for all cloud nodes...

Could this be an empty value for non-ipv6 nodes? so we could make the label unconditional but the value would be empty most of the time? Is there a better way?

nkinkade (commented):
> *[Reviewable](https://reviewable.io/reviews/m-lab/k8s-support/651)* status: 0 of 1 approvals obtained (waiting on @stephen-soltesz)

___
*[manage-cluster/add_k8s_virtual_node.sh, line 299 at r1](https://reviewable.io/reviews/m-lab/k8s-support/651#-MwXZsmXBU2zT5b4YsxZ:-MwXgTq49--118Bk2_F4:b-pd1x5f) ([raw file](https://github.com/m-lab/k8s-support/blob/655e336b5f956bda2997ef76b2f65650f4c8ff57/manage-cluster/add_k8s_virtual_node.sh#L299)):*

Google get's a C for consistently naming things.

We will need a way to plumb this value into an additional label for all cloud nodes...

Could this be an empty value for non-ipv6 nodes? so we could make the label unconditional but the value would be empty most of the time? Is there a better way?
</blockquote></details>

@stephen-soltesz: I've modified the script to unconditionally write the file `external-ipv6`.  If there is no IPv6, then the file will be empty. I believe it should, but will the `KeyValue` functionality of the flagx package handle this gracefully?

stephen-soltesz (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/m-lab/k8s-support/651#-:-MwXjD2aBXzcrabMCs16:bnfp4nl)* status: 0 of 1 approvals obtained

___
*[manage-cluster/add_k8s_virtual_node.sh, line 299 at r1](https://reviewable.io/reviews/m-lab/k8s-support/651#-MwXZsmXBU2zT5b4YsxZ:-MwXj1MB5iB6YEENscuk:byf6nq5) ([raw file](https://github.com/m-lab/k8s-support/blob/655e336b5f956bda2997ef76b2f65650f4c8ff57/manage-cluster/add_k8s_virtual_node.sh#L299)):*

@stephen-soltesz: I've modified the script to unconditionally write the file `external-ipv6`.  If there is no IPv6, then the file will be empty. I believe it should, but will the `KeyValue` functionality of the flagx package handle this gracefully?
</blockquote></details>

As long as the file there, and empty value should be no problem.

PR #659, opened 2022-03-07
Improves script for uprading k8s on virtual nodes

nkinkade:
The script previously assumed that all VMs needed to be upgraded. This
commits adds logic that will skip nodes that are already at the right
version.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/k8s-support/659)

### DIFF
--- a/manage-cluster/upgrade_k8s_virtual_nodes.sh
+++ b/manage-cluster/upgrade_k8s_virtual_nodes.sh
@@ -37,9 +37,23 @@ NODES=$(
       --template '{{range .items}}{{.metadata.name}}{{"\n"}}{{end}}'
 )
 
+for node in $NODES; do
+  UPGRADE_NODES=""
+  CURRENT_VERSION=$(
+    kubectl get node $node -o jsonpath='{.status.nodeInfo.kubeletVersion}{"\n"}'
+  )
+  lowest_version=$(
+    echo -e "${CURRENT_VERSION}\n${K8S_VERSION}" | sort --version-sort | head --lines 1
+  )
+
+  if [[ $lowest_version == $CURRENT_VERSION ]]; then
+    UPGRADE_NODES="${UPGRADE_NODES} ${node}"
+  fi
+done
+
 set +x
 echo -e "\n\n##### NODES TO BE UPGRADED #####"
-for node in $NODES; do
+for node in $UPGRADE_NODES; do
   echo $node
 done
 echo -e "\n"
@@ -58,7 +72,7 @@ if [[ "${keepgoing}" != "y" ]]; then
 fi
 set -x
 
-for node in $NODES; do
+for node in $UPGRADE_NODES; do
   # Replace the dots in the hostname with dashes.
   node="${node//./-}"
   # Determine the zone of the node.
--- a/manage-cluster/upgrade_k8s_virtual_nodes.sh
+++ b/manage-cluster/upgrade_k8s_virtual_nodes.sh
@@ -37,16 +37,15 @@ NODES=$(
       --template '{{range .items}}{{.metadata.name}}{{"\n"}}{{end}}'
 )
 
+UPGRADE_NODES=""
 for node in $NODES; do
-  UPGRADE_NODES=""
   CURRENT_VERSION=$(
     kubectl get node $node -o jsonpath='{.status.nodeInfo.kubeletVersion}{"\n"}'
   )
-  lowest_version=$(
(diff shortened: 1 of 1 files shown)


stephen-soltesz (changes requested):
> *[Reviewable](https://reviewable.io/reviews/m-lab/k8s-support/659)* status: 0 of 1 approvals obtained (waiting on @nkinkade)

___
*[manage-cluster/upgrade_k8s_virtual_nodes.sh, line 49 at r1](https://reviewable.io/reviews/m-lab/k8s-support/659#-MxaXqN8BqN-ULshS0md:-MxaXqN8BqN-ULshS0me:buhk2t3) ([raw file](https://github.com/m-lab/k8s-support/blob/46ce76d86f8677ce20543e5fec022cd54835c8c5/manage-cluster/upgrade_k8s_virtual_nodes.sh#L49)):*
> ```Shell
>   )
> 
>   if [[ $lowest_version == $CURRENT_VERSION ]]; then
> ```

if `K8S_VERSION` == `CURRENT_VERSION` then this condition would still be true (and an upgraded node would be added to `UPGRADE_NODES`). Is that intended? Is that a noop?

nkinkade (commented):
> *[Reviewable](https://reviewable.io/reviews/m-lab/k8s-support/659)* status: 0 of 1 approvals obtained (waiting on @stephen-soltesz)

___
*[manage-cluster/upgrade_k8s_virtual_nodes.sh, line 49 at r1](https://reviewable.io/reviews/m-lab/k8s-support/659#-MxaXqN8BqN-ULshS0md:-Mxa_0GxBaXrQD7rb50v:b-cj3yt4) ([raw file](https://github.com/m-lab/k8s-support/blob/46ce76d86f8677ce20543e5fec022cd54835c8c5/manage-cluster/upgrade_k8s_virtual_nodes.sh#L49)):*

if `K8S_VERSION` == `CURRENT_VERSION` then this condition would still be true (and an upgraded node would be added to `UPGRADE_NODES`). Is that intended? Is that a noop?
</blockquote></details>

Thank you for catching this bug. Geez, that is embarrassing. The script actually appeared to worked when I ran it a while ago, but only because of an additional bug and some naming chance. I'm setting `UPGRADE_NODES=""` _inside_ the loop, so it gets reset on every iteration. :/ Then, because kubectl returns an alphabetically sorted list of nodes `prometheus-platform-cluster` is the last in the list, meaning that this would have only offered up that node for possible upgrade, which was what I expected, but was, of course misleading. 

I believe the latest commit resolves this. It simplifies things. Rethinking it, there is no need to any sort of version sorting. CURRENT_VERSION should _never_ be greater than K8S_VERSION, so the script now assumes that if they differ then the VM needs upgrading. If CURRENT_VERSION is for some reason greater than K8S_VERSION, then the node should be effectively downgraded.

stephen-soltesz (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/m-lab/k8s-support/659#-:-Mxah-ef0mKe_LfeQ-HQ:bnfp4nl)* status: 0 of 1 approvals obtained (waiting on @nkinkade)

___
*[manage-cluster/upgrade_k8s_virtual_nodes.sh, line 49 at r1](https://reviewable.io/reviews/m-lab/k8s-support/659#-MxaXqN8BqN-ULshS0md:-MxafVvy76dyqfLAej7i:b-yq5lt7) ([raw file](https://github.com/m-lab/k8s-support/blob/46ce76d86f8677ce20543e5fec022cd54835c8c5/manage-cluster/upgrade_k8s_virtual_nodes.sh#L49)):*

Thank you for catching this bug. Geez, that is embarrassing. The script actually appeared to worked when I ran it a while ago, but only because of an additional bug and some naming chance. I'm setting `UPGRADE_NODES=""` _inside_ the loop, so it gets reset on every iteration. :/ Then, because kubectl returns an alphabetically sorted list of nodes `prometheus-platform-cluster` is the last in the list, meaning that this would have only offered up that node for possible upgrade, which was what I expected, but was, of course misleading. 

I believe the latest commit resolves this. It simplifies things. Rethinking it, there is no need to any sort of version sorting. CURRENT_VERSION should _never_ be greater than K8S_VERSION, so the script now assumes that if they differ then the VM needs upgrading. If CURRENT_VERSION is for some reason greater than K8S_VERSION, then the node should be effectively downgraded.
</blockquote></details>

Sometimes I feels like the longer I look at code the more bugs I see. I didn't even notice the reset of `UPGRADE_NODES=""` previously :)

PR #660, opened 2022-03-08
Allows kube-state-metrics to export select node label values

nkinkade:
By default, kube-state-metrics does not report of all possible resource labels. Instead, you must explicitly tell it which labels of which resources you want it to collect. This PR allows it to report on the `mlab/type` node label, which we can used to filter between "physical" and "virtual" nodes in our dashboard and queries.

Additionally:

* Configures the `PlatformCluster_TooManySwitchIfInErrors` alert to respect site GMX status.
* Configures boot disk type and size for virtual nodes instead using the default 10GB "standard" disk that GCE provisions.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/k8s-support/660)


stephen-soltesz (approved):
:lgtm:

> Reviewed 4 of 4 files at r1.
> *[Reviewable](https://reviewable.io/reviews/m-lab/k8s-support/660#-:-Mxjy9d_AnedwqXQf1-B:bnfp4nl)* status: 0 of 1 approvals obtained

Issue #637, opened 2022-01-20
Vector logs outage monitoring / alerting 


nkinkade:
This issue was originally created as https://github.com/m-lab/prometheus-support/issues/866:

Recently we discovered that some recent changes broke the vector reporting of container logs ot stackdriver.

More generally, we want to know that these logs are available in the event of some incident when they may be the only information we have remaining.

It would be worth while adding an alert on the logs being present so we know if it's not available in the future.

```
Jan 12 15:03:15.180 ERROR source{component_kind="source" component_name=kubernetes_logs component_type=kubernetes_logs}: vector::internal_events::kubernetes::instrumenting_watcher: Watch invocation failed. error=Other { source: Request { source: hyper::Error(Io, Os { code: 104, kind: ConnectionReset, message: "Connection reset by peer" }) } } rate_limit_secs=5
Jan 12 15:03:15.190 ERROR source{component_kind="source" component_name=kubernetes_logs component_type=kubernetes_logs}: vector::kubernetes::reflector: watcher error error=Request { source: hyper::Error(Io, Os { code: 104, kind: ConnectionReset, message: "Connection reset by peer" }) }
Jan 12 15:03:15.190 ERROR source{component_kind="source" component_name=kubernetes_logs component_type=kubernetes_logs}: vector::sources::kubernetes_logs: reflector process exited with an error error=Invocation { source: Request { source: hyper::Error(Io, Os { code: 104, kind: ConnectionReset, message: "Connection reset by peer" }) } }
Jan 12 15:03:15.203  INFO source{component_kind="source" component_name=kubernetes_logs component_type=kubernetes_logs}: vector::sources::kubernetes_logs: event processing loop completed gracefully
Jan 12 15:03:15.203  INFO source{component_kind="source" component_name=kubernetes_logs component_type=kubernetes_logs}: vector::sources::kubernetes_logs: file server completed gracefully
Jan 12 15:03:15.203  INFO source{component_kind="source" component_name=kubernetes_logs component_type=kubernetes_logs}: vector::sources::kubernetes_logs: done
```

nkinkade:
I believe we may be hitting this bug: https://github.com/vectordotdev/vector/issues/7401

Apparently that bug manifests itself when collection of `kubernetes_logs` fails to resume after an error. We are currently using a very old version of the vector container: timberio/vector:nightly-2020-10-12-debian

I will update Vector to a more recent version.

PR #665, opened 2022-03-29
Replaces Vector with fluent-bit

nkinkade:
Not long ago [we discovered that due to a probable bug in Vector](https://github.com/m-lab/k8s-support/issues/637), many pods were not pushing logs to Stackdriver. We upgraded Vector to a version with the bug supposed fix, and added a new alert to let us know when container logs for a node are not showing up in Stackdriver.

Lately, the new alerts for missing logs has started to fire in production and staging. It seems there is [some other unresolved bug in Vector](https://github.com/vectordotdev/vector/issues/8616). Logs are just to important for us to be missing, and for alerts to be firing all the time.

This PR replaces Vector with fluent-bit. We don't know that fluent-bit is necessarily better than Vector, but it is worth a shot. Vector seems very nice, and its configuration format is certainly far more elegant and powerful than that of fluent-bit, but as long as logs consistently make it to Stackdriver, then the extra complexity in configuration is worth it once.

_NOTE_: This PR also includes a random, unrelated fix to the script that creates virtual platform nodes. If the directory /etc/kubernetes/manifests does not exist, the kubelet spams its log with messages every few seconds complaining about this missing directory. The changes just creates the directory, if it doesn't already exist.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/k8s-support/665)


robertodauria (approved):
:lgtm: - Thanks. 

> Reviewed 5 of 6 files at r1.
> *[Reviewable](https://reviewable.io/reviews/m-lab/k8s-support/665#-:-MzOybu-0K3jgFgli-nN:bqhpokd)* status: 0 of 1 approvals obtained

PR #670, opened 2022-04-21
Sets the default flush value to 15s for fluent-bit

nkinkade:
See: https://github.com/m-lab/ops-tracker/issues/1561

The default flush value is 1s. We are beginning to run up against GCP
quotas and limits with regards to requests per minute to the logging
API. Hopefully this change will cause fluent-bit to flush local data to
the Stackdriver output less frequently, causing us to not hit quotas and
limits.

### DIFF
--- a/config/fluentbit/values.yaml.template
+++ b/config/fluentbit/values.yaml.template
@@ -90,6 +90,7 @@ extraVolumeMounts:
   readOnly: true
 - name: fluent-bit-etc
   mountPath: /fluent-bit/etc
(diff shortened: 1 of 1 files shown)


robertodauria (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/m-lab/k8s-support/670#-:-N0BzX9732wGhcOK3gyP:bnfp4nl)* status: 0 of 1 approvals obtained

PR #673, opened 2022-04-25
Adds explicit pod label for site type

nkinkade:
We have a need to be able to segregate metrics between experiment pods running in physical and virtual environments. This PR adds an explicit `measurementlab.net/type` label to all experiment pods. By default the value is "physical", but the ndt-virtual DaemonSet sets this label to "virtual". This [label value should propagate](https://github.com/m-lab/k8s-support/blob/master/config/prometheus/prometheus.yml#L130) to all experiment pod metrics. We can then use this metric label to separate metrics running on physical hardware from ones running in virtual environments.

On nodes, the equivalent label is currently `mlab/type`. In this case, I chose `measurementlab.net/type` because [the official k8s documentation on labels](https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#syntax-and-character-set) states:

> The prefix is optional. If specified, the prefix must be a DNS subdomain: a series of DNS labels separated by dots

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/k8s-support/673)

### DIFF
--- a/k8s/daemonsets/experiments/ndt-virtual.jsonnet
+++ b/k8s/daemonsets/experiments/ndt-virtual.jsonnet
@@ -35,6 +35,7 @@ exp.ExperimentNoIndex(expName, 'pusher-' + std.extVar('PROJECT_ID'), 'none', dat
         },
         labels+: {
           workload: expName + '-virtual',
+          measurementlab.net/type: 'virtual',
         },
       },
       spec+: {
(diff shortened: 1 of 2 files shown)


stephen-soltesz (approved):
Nice - these are carried to prometheus labels directly?

I see `measurementlab_net_type="virtual"` in prometheus... 

Our relabel configs could truncate that to `type=...`.

Consider when / how we may want to label things by platform type, e.g. "physical" for bare metal, "gcp" for virtual on gcp, maybe "aws" for virtual on aws in the future perhaps.

:lgtm:

> *[Reviewable](https://reviewable.io/reviews/m-lab/k8s-support/673#-:-N0agTYJ8kYLZ2Gk4JP1:b-6vusbu)* status: 0 of 1 approvals obtained

PR #674, opened 2022-04-26
Adds new alert for when node is unschedulable

nkinkade:
When a node is unschedulale for too long, fires an alert. This often happens when a node is flagged for reboot using Kured, and the node fails to come back up after 4h (the configured limit within Kured). When this happens, Kured moves on to the next node, leaving the other node in a cordoned state. Currently, we have no automated way of detecting this. I usually incidentally notice it and resolve the issue manually. This PR adds a new alert to warn us of this condition so that someone can manually intervene. I'm slightly concerned that this will be a noisy alert. We can wait and see and make adjustments as necessary.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/k8s-support/674)

### DIFF
--- a/config/prometheus/alerts.yml
+++ b/config/prometheus/alerts.yml
@@ -305,6 +305,32 @@ groups:
         Perhaps an API server node was incorrectly upgraded _not_ using the
         ugprade script and not all components got updated?
 
+  # If any node is unschedulable for too long, fire an alert, unless the node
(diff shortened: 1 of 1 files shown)


robertodauria (approved):
:lgtm: -- Thanks!

> *[Reviewable](https://reviewable.io/reviews/m-lab/k8s-support/674#-:-N0fJMKI5pL4WEUOU5RB:b-kejgc9)* status: 0 of 1 approvals obtained

PR #678, opened 2022-04-28
Bumps fluent-bit version to v1.9.3-debug

nkinkade:
v1.9.3 is the latest release, and we use the debug image, which is based
on debian:bullseye-slim base image. Since we are new to using
fluent-bit, when problems arise we really need to be able to get into
the container to debug. The default image is based on "distroless",
which has absolutely no way to interact with the container. There is no
harm in running the debian-slim image for now. When we are confident
that fluent-bit is running flawlessly without any issues, then we can
perhaps revert to the non-debug image.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/k8s-support/678)


stephen-soltesz (approved):
> *[Reviewable](https://reviewable.io/reviews/m-lab/k8s-support/678)* status: 0 of 1 approvals obtained

PR #700, opened 2022-07-13
Sets some limits for the Stackdriver output plugin

nkinkade:
We are having some issues with fluent-bit where a flaky or down
Internet connection causes it to stop sending log data, and it never
recovers. This commit introduces a few new configuration settings which
are aimed to workaround this probable bug in fluent-bit, which are also
probably sensible in any case. We don't want fluent-bit to fill up the
filesystem, and in the case of extended outages we don't need to worry
about trying to send data forever. It is okay for us to lose some log
data.

The motivation for this PR is to maybe workaround [an issue we are seeing with fluent-bit](https://github.com/m-lab/ops-tracker/issues/1594) when it encounters and Internet outage or a flaky connection.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/k8s-support/700)

### DIFF
--- a/config/fluentbit/values.yaml.template
+++ b/config/fluentbit/values.yaml.template
@@ -116,6 +116,14 @@ initContainers:
         namespace ${NODE_NAME}
         node_id ${NODE_NAME}
         export_to_project_id {{PROJECT}}
+        # Don't try to push a chunk forever.
+        Retry_Limit 10
+        # Default timeout is 10s. Increase this a bit to workaround a slow or
+        # flaky network connection.
+        net.connect_timeout 60
+        # Don't continue to fill up the filesystem with buffered chunks when
+        # pushing logs to Stackdriver isn't working for some reason.
+        storage.total_limit_size 1G
     EOF
   env:
   - name: NODE_NAME


stephen-soltesz (approved):
> *[Reviewable](https://reviewable.io/reviews/m-lab/k8s-support/700)* status: 0 of 1 approvals obtained (waiting on @nkinkade)

___
*[`config/fluentbit/values.yaml.template` line 126 at r1](https://reviewable.io/reviews/m-lab/k8s-support/700#-N6tZyPf3h8t0mzjrsqv:-N6tZyPf3h8t0mzjrsqw:b5s2bm) ([raw file](https://github.com/m-lab/k8s-support/blob/dfaddb096e92bfbfff71fbc1b390fd70caacdfbc/config/fluentbit/values.yaml.template#L126)):*
> ```
>         # Don't continue to fill up the filesystem with buffered chunks when
>         # pushing logs to Stackdriver isn't working for some reason.
>         storage.total_limit_size 1G
> ```

Conversationally, we talked about RAM buffering. Does this setting still apply to on disk storage? (that's fine too, just clarifying).

If this is still an issue, we could also try `Retry_limit no_retries` - https://docs.fluentbit.io/manual/administration/scheduling-and-retries#configuring-retries to be even more aggressive about discarding logs that it can't upload.

Similar to how we regard prometheus metrics as advisory (not ground truth) and ephemeral (deleted after 120days), I think of logs data as being equally advisory and best effort, if some get lost from transient events, it's okay. As you've observed too, reducing our logging output to reduce the stress on the collection process would also help all factors.

nkinkade (commented):
> *[Reviewable](https://reviewable.io/reviews/m-lab/k8s-support/700)* status: 0 of 1 approvals obtained (waiting on @stephen-soltesz)

___
*[`config/fluentbit/values.yaml.template` line 126 at r1](https://reviewable.io/reviews/m-lab/k8s-support/700#-N6tZyPf3h8t0mzjrsqv:-N6tbCpn2I3ExNzZu5cK:b76s6mz) ([raw file](https://github.com/m-lab/k8s-support/blob/dfaddb096e92bfbfff71fbc1b390fd70caacdfbc/config/fluentbit/values.yaml.template#L126)):*

Conversationally, we talked about RAM buffering. Does this setting still apply to on disk storage? (that's fine too, just clarifying).

If this is still an issue, we could also try `Retry_limit no_retries` - https://docs.fluentbit.io/manual/administration/scheduling-and-retries#configuring-retries to be even more aggressive about discarding logs that it can't upload.

Similar to how we regard prometheus metrics as advisory (not ground truth) and ephemeral (deleted after 120days), I think of logs data as being equally advisory and best effort, if some get lost from transient events, it's okay. As you've observed too, reducing our logging output to reduce the stress on the collection process would also help all factors.
</blockquote></details>

You are right, we did talk about disabling filesystem buffering altogether to see if it would work around this issue. A search for similar fluent-bit problems [turns up](https://github.com/fluent/fluent-bit/issues/3581) [quite a lot](https://github.com/fluent/fluent-bit/issues/3328) of [Github issues](https://github.com/fluent/fluent-bit/issues/3308). In one of them, [someone had already tried disabling filesystem buffering](https://github.com/fluent/fluent-bit/issues/2950#issuecomment-762818927), but noted that the problem persisted, so I didn't bother trying to go that route in this PR.

The changes in this PR may not help this specific issue we are seeing, but I think they cannot hurt, and _might_ help in some cases.

PR #702, opened 2022-07-19
Fixes fluentbit alert + removes custom DNS config

nkinkade:
The fluentbit DaemonSet named changed from "fluent-bit" to "fluentbit", but the alert was mistakenly never updated.

We noticed DNS lookup failures in some problem fluentbit containers, and wondered whether using Google public DNS servers would help, much like we do in NDT containers. However, we failed to take into account that fluentbit needs to resolve cluster-local addresses, which are private and not resolvable by public nameservers. This PR reverts that change.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/k8s-support/702)

### DIFF
--- a/config/fluentbit/values.yaml.template
+++ b/config/fluentbit/values.yaml.template
@@ -74,11 +74,6 @@ config:
         HTTP_Listen  0.0.0.0
         HTTP_PORT    2020
         storage.path /var/lib/fluent-bit
-dnsConfig:
-  nameservers:
-  - 8.8.8.8
-  - 8.8.4.4
-dnsPolicy: None
 env:
 - name: GOOGLE_APPLICATION_CREDENTIALS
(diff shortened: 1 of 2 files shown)


stephen-soltesz (approved):
Hmm, is this a DNS lookup request limit? Do we have other examples of services doing both cluster and public DNS lookups?

> *[Reviewable](https://reviewable.io/reviews/m-lab/k8s-support/702#-:-N7MvcWR6MqHyuNO3pkl:blgq7iq)* status: 0 of 1 approvals obtained

nkinkade (commented):
Good question, and idea. It is certainly something worth looking into. I will investigate.

> *[Reviewable](https://reviewable.io/reviews/m-lab/k8s-support/702#-:-N7MxqAqE8H1r9smyKd9:btupqwz)* status: 0 of 1 approvals obtained

PR #707, opened 2022-07-27
Sets ndots=2 in dnsPolicy for fluentbit and disco DaemonSets + fix flannel

nkinkade:
Currently, and by default in k8s, pods have a resolver policy of `ndots:5`, which means that if a host name has fewer than 5 dots, the resolver will assume it is a relative name, and will proceed to append each search domain to the name and do a lookup. For example, if the resolver sees a name like "logging.googleapis.com" (which only has 2 dots), it will first do lookups of things like "logging.googleapis.com.default.svc.cluster.local" (which will fail), and then "logging.googleapis.com.svc.cluster.local", and so on, until it has exhausted all search domains, _then_ and finally it will issue a final FQDN recursive search for "logging.googleapis.com." (dot at end), which will succeed. This means that just to lookup "logging.googleapis.com" will always entail at least 4 DNS lookups before an IP is received.

On the M-Lab platform we mostly do resolution using Google's public DNS servers, and very few intra cluster lookup are done. It turns out that fluentbit and disco are two services that mostly use the cluster's CoreDNS instances for resolvers. This PR sets `ndots:2` for both of those DaemonSets. This means that cluster-local lookups may take require two queries, but observation shows that the vast majority of lookups these services are doing are for names like "oauth2.googleapis.com" and "logging.googleapis.com", etc. This change is a pretty big optimization for these services.

Consider this dashboard panel for CoreOS in production:

Screenshot from 2022-07-27 10-52-39

The NXDOMAIN responses are triple the rate of NOERROR. Compare that with sandbox, which contains this optimization:

Screenshot from 2022-07-27 10-52-56

You can see how things used to be yesterday, with NXDOMAINs prominent. But with this change, the NXDOMAIN responses have dropped to nearly 0, and now NOERROR responses comprise the majority of responses.

**Additionally**: a bug was introduced in a previous PR, which had flannel containers looking for the container image in the wrong place. This PR fixes that too.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/k8s-support/707)

### DIFF
--- a/config/fluentbit/values.yaml.template
+++ b/config/fluentbit/values.yaml.template
@@ -74,6 +74,10 @@ config:
         HTTP_Listen  0.0.0.0
         HTTP_PORT    2020
         storage.path /var/lib/fluent-bit
+dnsConfig:
+  options:
+  - name: ndots
+    value: 2
 env:
 - name: GOOGLE_APPLICATION_CREDENTIALS
   value: /etc/fluentbit/keys/fluentbit.json
--- a/config/fluentbit/values.yaml.template
+++ b/config/fluentbit/values.yaml.template
@@ -78,6 +78,10 @@ dnsConfig:
   options:
   - name: ndots
     value: "2"
+  searches:
+  - cluster.local
+  - svc.cluster.local
+  - default.svc.cluster.local
 env:
 - name: GOOGLE_APPLICATION_CREDENTIALS
   value: /etc/fluentbit/keys/fluentbit.json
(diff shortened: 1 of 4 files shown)


nkinkade:
I added one more commit, setting `net.dns.mode=TCP`, which other users have indicated fully or partially resolves DNS resolution issues.

stephen-soltesz (approved):
Thank you.


:lgtm:

> Reviewed 4 of 4 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/m-lab/k8s-support/707#-:-N80e1ck-Ox2ZrBXcoLr:b-h0y7g4)* status: 0 of 1 approvals obtained

PR #701, opened 2022-07-18
Better support for physical/virtual dashboards + updates to cAdvisor config 

nkinkade:
The majority of changes in this PR are to allow better support for switching on virtual/physical in various Grafana dashboards, mostly be updating recording rules, but also by adding a `site` label to more metrics.

While working on the dashboard updates, we also realized that cAdvisor was exporting far too many metrics we don't use, with far too many labels. This PR also reconfigures cAdvisor to export fewer metrics with fewer labels.

Additionally this PR causes kube-state-metrics to run on the Prometheus VM. Currently, the nodeSelector for the pod is `mlab/type="virtual"`, which in the past would have ensured that it was running on one of the control plane nodes or on the Prometheus node. Today, we have virtual sites, and I found that kube-state-metrics was running on random virtual nodes, which didn't seem desirable.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/k8s-support/701)

PR #708, opened 2022-07-28
Update node_exporter to v1.3.1 + reorganize flags

nkinkade:
In a similar spirit to [what we did with cAdvisor](https://github.com/m-lab/k8s-support/pull/701), this PR attempts to reduce the number of node_exporter collectors which we have enabled to reduce the number of unused timeseries in Prometheus. Currently, we accept the default list of enabled collectors, and then explicitly disable the ones we don't want. This PR reverses that by disabling defaults and enabling only the collectors that expose metrics that we use today. This helps to clean up the flags a good deal.

Also, I removed these flags, since their default values in the collector were [either identical](https://github.com/prometheus/node_exporter/blob/v1.3.1/collector/filesystem_linux.go#L36) or [almost identical](https://github.com/prometheus/node_exporter/blob/v1.3.1/collector/filesystem_linux.go#L37):
* --collector.filesystem.mount-points-exclude
* --collector.filesystem.fs-types-exclude

I disabled the `netstat` and `systemd` collectors, since we don't seem to use any metrics from the former, and the latter is broken because of some default container AppArmor policy likely set by containerd.
 
By recursive grepping of both this repo and the prometheus-support repo, I identified the following node_exporter metrics that we use, organized by the collector that exposes them. I verified that they all still exist with the new configuration. Are there metrics I missed, orplaces I didn't look that might use metrics from node_exporter?

cpu
===
node_cpu_seconds_total

diskstats
=========
node_disk_io_time_seconds_total
node_disk_read_bytes_total
node_disk_written_bytes_total

edac
====
node_edac_uncorrectable_errors_total
node_edac_correctable_errors_total

filesystem
==========
node_filesystem_avail_bytes
node_filesystem_size_bytes
node_filesystem_free_bytes
node_filesystem_files_free

hwmon
=====
node_hwmon_temp_celsius

loadavg
=======
node_load1
node_load5
node_load15

meminfo
=======
node_memory_MemTotal_bytes
node_memory_MemFree_bytes
node_memory_Cached_bytes
node_memory_Buffers_bytes
node_memory_Active_bytes
node_memory_MemAvailable_bytes

netdev
=====
node_network_receive_bytes_total
node_network_transmit_bytes_total

processes
=========
node_processes_pids

stat
====
node_boot_time_seconds
node_procs_running

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/k8s-support/708)


stephen-soltesz (approved):
... there is good stuff in the netstat metrics but we've never realized it. and, I'm not sure the current config was reporting everything or just the host environment... I don't object to the removal but this could be a fund side project to tease out the ListenOverflow and ListenDrop events from the ndt container.. That's a form of capacity utilization that we don't otherwise monitor/track today and a source of potential failure for users.

https://prometheus-platform-cluster.mlab-oti.measurementlab.net/graph?g0.expr=node_netstat_TcpExt_ListenDrops%7Bmachine%3D~%22mlab.......*%22%7D%20%3E%200&g0.tab=0&g0.stacked=0&g0.show_exemplars=0&g0.range_input=1w

Thank you for the careful accounting of the metrics currently in use.

:lgtm:

> *[Reviewable](https://reviewable.io/reviews/m-lab/k8s-support/708#-:-N85JNgd-IQHiWgS-xiz:b-j3831m)* status: 0 of 1 approvals obtained

PR #710, opened 2022-07-29
Adds a new script to virtual nodes to configure fq qdisc + update revtr 

nkinkade:
We are now wanting to export a metric via node_exporter's textfile collector that will indicate whether the fq qdisc was properly
configured on a machine. The previous systemd service baked into virtual machines was very simple and could not create this new metric file. This commit creates a new script in the node cloud-init configuration which will configure the fq qdisc and create a metric file. The systemd service now simply calls this script. It's unfortunate that we need separate scripts to do this job on physical and virtual nodes, but that is how things need to be for now, as far as I can tell.

Also, this PR bumps the revtr container image to version v0.2.3.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/k8s-support/710)

### DIFF
--- a/k8s/daemonsets/experiments/revtr.jsonnet
+++ b/k8s/daemonsets/experiments/revtr.jsonnet
@@ -7,7 +7,7 @@ exp.Experiment('revtr', 3, 'pusher-' + std.extVar('PROJECT_ID'), 'none', []) + {
         containers+: [
           {
             name: 'revtrvp',
-            image: 'measurementlab/revtrvp:v0.2.1',
+            image: 'measurementlab/revtrvp:v0.2.2',
             args: [
               '/root.crt',
               '/plvp.config',
--- a/k8s/daemonsets/experiments/revtr.jsonnet
+++ b/k8s/daemonsets/experiments/revtr.jsonnet
@@ -7,7 +7,7 @@ exp.Experiment('revtr', 3, 'pusher-' + std.extVar('PROJECT_ID'), 'none', []) + {
         containers+: [
           {
             name: 'revtrvp',
-            image: 'measurementlab/revtrvp:v0.2.2',
+            image: 'measurementlab/revtrvp:v0.2.3',
             args: [
               '/root.crt',
               '/plvp.config',
--- a/manage-cluster/cloud-config_node.yml
+++ b/manage-cluster/cloud-config_node.yml
@@ -53,7 +53,7 @@ write_files:
     #!/bin/bash
     # Determine the default/primary network interface of the VM.
     IFACE=$(ip -o -4 route show default | awk '{print $5}')
-    # For vitual machines we set the fq qdisc maxrate to 10g, here expressed at
+    # For vitual machines we set the fq qdisc maxrate to 10g, here expressed as
     # bytes per second.
     MAXRATE="1250000000"
     # This script writes out a Prometheus metric file which will be collected by the
@@ -65,7 +65,7 @@ write_files:
     echo -n "node_configure_qdisc_success " > $METRIC_FILE_TEMP
     # Append the passed status code to the temporary metric file, overwrite the
     # metric file with the temp metric file, and make the metric file world readable.
-    function overwrite_metric_file {
+    function write_metric_file {
       local status=$1
       echo "$status" >> $METRIC_FILE_TEMP
       mv $METRIC_FILE_TEMP $METRIC_FILE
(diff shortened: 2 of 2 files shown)


stephen-soltesz (approved):
One clarifying question, but otherwise :lgtm:

> *[Reviewable](https://reviewable.io/reviews/m-lab/k8s-support/710#-:-N8ALXhu7m7Z123GZOgg:b-4j3rri)* status: 0 of 1 approvals obtained (waiting on @nkinkade)

___
*[`manage-cluster/cloud-config_node.yml` line 74 at r1](https://reviewable.io/reviews/m-lab/k8s-support/710#-N8AKGx18TLE4lbzcYV-:-N8AKGx18TLE4lbzcYV0:b-ujx9oe) ([raw file](https://github.com/m-lab/k8s-support/blob/142eead110c45860ad768a178019a11fdac9bd35/manage-cluster/cloud-config_node.yml#L74)):*
> ```YAML
>       chmod 644 $METRIC_FILE
>     }
>     tc qdisc replace dev $IFACE root fq maxrate "${MAXRATE}bps"
> ```

Just clarifying: tc uses `bps` to abbreviate "bytes per second"?

If not this could be a unit mismatch with the described value of `MAXRATE` above.

stephen-soltesz (commented):
> *[Reviewable](https://reviewable.io/reviews/m-lab/k8s-support/710)* status: 0 of 1 approvals obtained (waiting on @nkinkade)

___
*[`manage-cluster/cloud-config_node.yml` line 70 at r1](https://reviewable.io/reviews/m-lab/k8s-support/710#-N8ANrQg-u9b-IxBLNkr:-N8ANrQg-u9b-IxBLNks:b74xj70) ([raw file](https://github.com/m-lab/k8s-support/blob/142eead110c45860ad768a178019a11fdac9bd35/manage-cluster/cloud-config_node.yml#L70)):*
> ```YAML
>     function overwrite_metric_file {
>       local status=$1
>       echo "$status" >> $METRIC_FILE_TEMP
> ```

I noticed this in the other PR too - `>>` is an append operator.

nkinkade (commented):
> *[Reviewable](https://reviewable.io/reviews/m-lab/k8s-support/710)* status: 0 of 1 approvals obtained (waiting on @stephen-soltesz)

___
*[`manage-cluster/cloud-config_node.yml` line 70 at r1](https://reviewable.io/reviews/m-lab/k8s-support/710#-N8ANrQg-u9b-IxBLNkr:-N8AWM1ABS2dfcJIH3iH:b-5pzqj6) ([raw file](https://github.com/m-lab/k8s-support/blob/142eead110c45860ad768a178019a11fdac9bd35/manage-cluster/cloud-config_node.yml#L70)):*

I noticed this in the other PR too - `>>` is an append operator.
</blockquote></details>

See the other PR for the explanation. It's just appending the value to the end of the existing temporary metric file.
___
*[`manage-cluster/cloud-config_node.yml` line 74 at r1](https://reviewable.io/reviews/m-lab/k8s-support/710#-N8AKGx18TLE4lbzcYV-:-N8AWWkjDm7FFPTMnzJO:b-gn8rm8) ([raw file](https://github.com/m-lab/k8s-support/blob/142eead110c45860ad768a178019a11fdac9bd35/manage-cluster/cloud-config_node.yml#L74)):*

Just clarifying: tc uses `bps` to abbreviate "bytes per second"?

If not this could be a unit mismatch with the described value of `MAXRATE` above.
</blockquote></details>

Yeah, this is confusing, because it is somewhat contrary to the way you usually see rates abbreviate but for some reason `tc` uses `bps` to mean "bytes per second":

https://man7.org/linux/man-pages/man8/tc.8.html#PARAMETERS

The reason I chose to use it here, instead of the slightly more intuitive `10gbit` is that `tc` stores the rate internally as a 32bit unsigned integer as "bps", so when I query with `tc qdisc show`, the value that comes back is in "bps". So using that value instead makes comparison easier when checking to be sure the value was set right.

stephen-soltesz (commented):
> *[Reviewable](https://reviewable.io/reviews/m-lab/k8s-support/710)* status: 0 of 1 approvals obtained

___
*[`manage-cluster/cloud-config_node.yml` line 74 at r1](https://reviewable.io/reviews/m-lab/k8s-support/710#-N8AKGx18TLE4lbzcYV-:-N8AZqVv7urncvjzUuaM:b-cqttwu) ([raw file](https://github.com/m-lab/k8s-support/blob/142eead110c45860ad768a178019a11fdac9bd35/manage-cluster/cloud-config_node.yml#L74)):*

Yeah, this is confusing, because it is somewhat contrary to the way you usually see rates abbreviate but for some reason `tc` uses `bps` to mean "bytes per second":

https://man7.org/linux/man-pages/man8/tc.8.html#PARAMETERS

The reason I chose to use it here, instead of the slightly more intuitive `10gbit` is that `tc` stores the rate internally as a 32bit unsigned integer as "bps", so when I query with `tc qdisc show`, the value that comes back is in "bps". So using that value instead makes comparison easier when checking to be sure the value was set right.
</blockquote></details>

Thank you. All good :+1:

PR #711, opened 2022-08-02
Auto-reboot for virtual nodes + updated alerts

nkinkade:
This PR adds the appropriate systemd service and timer, as well as a shell script, which should cause the virtual nodes to be rebooted when uptime exceeds 60d. Also included is a change which should cause systemd to wait for 180s before shutting down or rebooting the node. These features have already existed for physical nodes for quite a while, and this PR now brings this same functionality to virtual nodes. 

I also took this opportunity to modify the old alert `PlatformCluster_TooManyKubeletOrKernelVersions`, and break it apart into two new alerts:

* `PlatformCluster_TooManyOSVersions`: this alert now takes the place of the part of the old one that checked kernel versions. With the introduction of virtual nodes onto the platform, we can no longer guarantee that exactly one kernel version is running. Kernel version was probably too specific to begin with. This alert now just checks to be sure that all nodes are running the same Ubuntu OS release (e.g., "20.04").
* `PlatformCluster_TooManyKubeletVersions`: Like before, this alert just makes sure that there is only a single kubelet version on all nodes.

I have also bumped up the time threshold from 7d to 30d for these alerts. Sometimes upgrading kubernetes or the base OS can take some time, and this change gives us plenty of time for those things to happen with no rush.

Additionally, this PR introduces a small operator script which can facilitate updating all virtual nodes. It can only introduce changes that are part of the cloud-init system. The script overwrites the `user-data` metadata field of a virtual node (which is what cloud-init uses for it's config), then "clean"s the cloud-init directory on the node, such that on the next reboot cloud-init will reprocess the newly uploaded config, as if this was a branch new VM. This script perhaps isn't ideal, but it is a small stopgap until we can find a more permanent solution like Terraform.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/k8s-support/711)

### DIFF
--- a/config/prometheus/alerts.yml
+++ b/config/prometheus/alerts.yml
@@ -240,7 +240,7 @@ groups:
       dashboard: https://grafana.mlab-oti.measurementlab.net/d/_fugwnWZk/ops-tactical-and-sre-overview
 
   # Too many OS releases on the platform for too long. This could happen, for
-  # example, if a new epoxy-images version was onlhy partially deployed to the
+  # example, if a new epoxy-images version was only partially deployed to the
   # platform.
   - alert: PlatformCluster_TooManyOSVersions
     expr: |
--- a/config/prometheus/alerts.yml
+++ b/config/prometheus/alerts.yml
@@ -239,40 +239,42 @@ groups:
       description: https://github.com/m-lab/ops-tracker/wiki/Alerts-&-Troubleshooting#platformhardware_machineuptoolong
       dashboard: https://grafana.mlab-oti.measurementlab.net/d/_fugwnWZk/ops-tactical-and-sre-overview
 
-  # Too many kubelet or kernel versions on the platform for too long. This is
-  # an indication that a rolling reboot has stalled.
-  - alert: PlatformCluster_TooManyKubeletOrKernelVersions
+  # Too many OS releases on the platform for too long. This could happen, for
+  # example, if a new epoxy-images version was onlhy partially deployed to the
+  # platform.
+  - alert: PlatformCluster_TooManyOSVersions
     expr: |
-      count(
-        count by (kubelet_version) (
-          kube_node_info and on(node) kube_node_labels{label_mlab_type="physical"}
+      count (
+        count by (os_release) (
+          label_replace(kube_node_info, "os_release", "$1", "os_image", "Ubuntu ([0-9]{2}.[0-9]{2}).*")
         )
       ) > 1
-      or
-      count(
-        count by (kernel_version) (
-          kube_node_info and on(node) kube_node_labels{label_mlab_type="physical"}
-        )
-      ) > 1
-    for: 7d
+    for: 30d
     labels:
       repo: ops-tracker
       severity: ticket
       cluster: platform
     annotations:
-      summary: Too many kubelet or kernel versions in the cluster for too long.
-      description: >
-        There has been more than one kubelet or kernel version running in the
-        platform cluster for too long. Probably the only reason there should be
-        more than one version in the cluster is when kubernetes and/or CoreOS
-        has been upgraded in epoxy-images and a rolling reboot has not
-        completed to apply those changes. Since CLUO will only reboot a single
-        node at a time, stalls of rolling reboots can be easily caused by a
-        node in a NotReady state.  Check for nodes in a NotReady state, and if
-        necessary delete them from the cluster so the rolling reboot can
-        continue. Looks at the log of the update-operator pod in the
-        reboot-coordinator namespace to see which node it may be stuck on.
-      dashboard: https://grafana.mlab-oti.measurementlab.net/d/tZHLFQRZk/k8s-workload-overview
+      summary: Too many OS versions in the cluster for more than 30d.
+      description: https://github.com/m-lab/ops-tracker/wiki/Alerts-&-Troubleshooting#platformcluster_toomanyosversions 
+      dashboard: https://grafana.mlab-oti.measurementlab.net/d/_fugwnWZk/ops-tactical-and-sre-overview?orgId=1&viewPanel=22
+
+  # Too many kubelet versions on the platform for too long. This could happen, for
+  # example, if an upgrade to kubernetes is only partially completed.
+  - alert: PlatformCluster_TooManyKubeletVersions
+    expr: |
+      count(
+        count by (kubelet_version) (kube_node_info)
+      ) > 1
+    for: 30d
+    labels:
+      repo: ops-tracker
+      severity: ticket
+      cluster: platform
+    annotations:
+      summary: Too many kubelet versions in the cluster for more than 30d.
+      description: https://github.com/m-lab/ops-tracker/wiki/Alerts-&-Troubleshooting#platformcluster_toomanykubeletversions
+      dashboard: https://grafana.mlab-oti.measurementlab.net/d/_fugwnWZk/ops-tactical-and-sre-overview?orgId=1&viewPanel=22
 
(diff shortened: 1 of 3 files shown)


stephen-soltesz (approved):
I think my comments are only observations and clarifying questions. One potential refinement. :lgtm:

And, thank you for your consistently meticulous PR descriptions. They are always excellent.

> Reviewed all commit messages.
> *[Reviewable](https://reviewable.io/reviews/m-lab/k8s-support/711#-:-N8VGv9UAv76ZDPGmq-E:b-pedijr)* status: 0 of 1 approvals obtained (waiting on @nkinkade)

___
*[`manage-cluster/cloud-config_node.yml` line 113 at r1](https://reviewable.io/reviews/m-lab/k8s-support/711#-N8VB5FD3KCABOx6HKRs:-N8VB5FD3KCABOx6HKRt:b-ool8cv) ([raw file](https://github.com/m-lab/k8s-support/blob/ba5050ed9616bffa8880a3743f4a66019ebe4ad7/manage-cluster/cloud-config_node.yml#L113)):*
> ```YAML
>   content: |
>     [Manager]
>     DefaultTimeoutStopSec=180
> ```

FYI: When Locate v2 has migrated fully to the heartbeat service health status, this delay can be significantly shortened.
___
*[`manage-cluster/update_virtual_nodes_config.sh` line 8 at r1](https://reviewable.io/reviews/m-lab/k8s-support/711#-N8VELaIAe07Bej3UAsX:-N8VELaIAe07Bej3UAsY:bf3hixd) ([raw file](https://github.com/m-lab/k8s-support/blob/ba5050ed9616bffa8880a3743f4a66019ebe4ad7/manage-cluster/update_virtual_nodes_config.sh#L8)):*
> ```Shell
> # NOTE: Running `cloud-init clean` causes the host's ssh private keys to be
> # regenerated, which will causes errors about the host key changing the next
> # time you try to ssh into the VM.
> ```

Hmm, does your recent work for ssh key certificates apply to the virtual nodes as well? Could it?
___
*[`manage-cluster/update_virtual_nodes_config.sh` line 33 at r1](https://reviewable.io/reviews/m-lab/k8s-support/711#-N8V86-38QUTig--pW2f:-N8V86-38QUTig--pW2g:bwd4neb) ([raw file](https://github.com/m-lab/k8s-support/blob/ba5050ed9616bffa8880a3743f4a66019ebe4ad7/manage-cluster/update_virtual_nodes_config.sh#L33)):*
> ```Shell
>   # cloud-init to configure the VM.
>   gcloud compute instances add-metadata $vm \
>     --metadata-from-file="user-data=${CLOUDINIT_FILE}" \
> ```

Just clarifying: is this metadata removed after a successful cloud init run? Or, this simply updates the content of user-data and the `cloud-init clean` resets vm-local state that controls whether this user-data content is executed? That seems like that's it.
___
*[`manage-cluster/update_virtual_nodes_config.sh` line 41 at r1](https://reviewable.io/reviews/m-lab/k8s-support/711#-N8VGgAM8frtEFMf-imq:-N8VGgAM8frtEFMf-imr:b-q6wqj6) ([raw file](https://github.com/m-lab/k8s-support/blob/ba5050ed9616bffa8880a3743f4a66019ebe4ad7/manage-cluster/update_virtual_nodes_config.sh#L41)):*
> ```Shell
>   # cloud-init configuration data, causing cloud-init to run all modules on the
>   # next boot. Pipe all output to /dev/null, since gcloud gets upset when the
>   # SSH connection is disconnected by the reboot.
> ```

`shutdown -r +1` may allow the ssh to exit cleanly without ignoring other potential error messages from the connection.

nkinkade (commented):
Thank you for the review. I believe I have addressed your questions and suggestions. PTAL?

> *[Reviewable](https://reviewable.io/reviews/m-lab/k8s-support/711#-:-N8VLUT95alT2giebdUo:b-bt3xbb)* status: 0 of 1 approvals obtained (waiting on @stephen-soltesz)

___
*[`manage-cluster/cloud-config_node.yml` line 113 at r1](https://reviewable.io/reviews/m-lab/k8s-support/711#-N8VB5FD3KCABOx6HKRs:-N8VILDVDuhN45O7Bj0V:b-f1bxef) ([raw file](https://github.com/m-lab/k8s-support/blob/ba5050ed9616bffa8880a3743f4a66019ebe4ad7/manage-cluster/cloud-config_node.yml#L113)):*

FYI: When Locate v2 has migrated fully to the heartbeat service health status, this delay can be significantly shortened.
</blockquote></details>

Good point. It will make working with experiment pods much easier and faster.
___
*[`manage-cluster/update_virtual_nodes_config.sh` line 8 at r1](https://reviewable.io/reviews/m-lab/k8s-support/711#-N8VELaIAe07Bej3UAsX:-N8VITQR3_YcAx6ItZeM:b-sv7np1) ([raw file](https://github.com/m-lab/k8s-support/blob/ba5050ed9616bffa8880a3743f4a66019ebe4ad7/manage-cluster/update_virtual_nodes_config.sh#L8)):*

Hmm, does your recent work for ssh key certificates apply to the virtual nodes as well? Could it?
</blockquote></details>

It doesn't yet apply to virtual nodes. It is currently only supported by a special part of the stage3_ubuntu epoxy-images build, which leverages its access to secrets in the Secret Manager from the build:

https://github.com/m-lab/epoxy-images/blob/master/setup_stage3_ubuntu.sh#L215

I'm sure we can figure out some way to sign SSH keys using the CA private key, but it will likely require some backbending. Possibly it could happen when the VM is created, and leverage the operator's access to secrets and the VM.
___
*[`manage-cluster/update_virtual_nodes_config.sh` line 33 at r1](https://reviewable.io/reviews/m-lab/k8s-support/711#-N8V86-38QUTig--pW2f:-N8VJW4sE1Elnw6B_Fb5:befe5b4) ([raw file](https://github.com/m-lab/k8s-support/blob/ba5050ed9616bffa8880a3743f4a66019ebe4ad7/manage-cluster/update_virtual_nodes_config.sh#L33)):*

Just clarifying: is this metadata removed after a successful cloud init run? Or, this simply updates the content of user-data and the `cloud-init clean` resets vm-local state that controls whether this user-data content is executed? That seems like that's it.
</blockquote></details>

Yes, that is right. The `user-data` metadata field gets populated when the VM is created, and it remain there for the life of the VM. When cloud-init runs on the instance it looks into `/var/lib/cloud` to gain state, and generally only runs most modules once per instance. Running `cloud-init clean` on the VM basically empties out the content of `/var/lib/cloud`, and when `cloud-init` notices this on boot, it acts if this is the first boot of the machine and runs all modules located in its config, which is located on the metadata server under the field `user-data`.
___
*[`manage-cluster/update_virtual_nodes_config.sh` line 41 at r1](https://reviewable.io/reviews/m-lab/k8s-support/711#-N8VGgAM8frtEFMf-imq:-N8VLAgP9QawfpT-mf1P:bj8ebs4) ([raw file](https://github.com/m-lab/k8s-support/blob/ba5050ed9616bffa8880a3743f4a66019ebe4ad7/manage-cluster/update_virtual_nodes_config.sh#L41)):*

`shutdown -r +1` may allow the ssh to exit cleanly without ignoring other potential error messages from the connection.
</blockquote></details>

Good idea. I implemented this suggestion.

stephen-soltesz (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/m-lab/k8s-support/711#-:-N8VMK1V7ICTNS11kTKM:bnfp4nl)* status: 0 of 1 approvals obtained (waiting on @stephen-soltesz)

PR #719, opened 2022-08-11
Don't fire qdisc alert if the node's status is NotReady

nkinkade:
If the node is in a NotReady state, then it is likely down, meaning that
the qdisc metric will not exist, since Prometheus can't scrape
node_exporter on the node.

- - -
This change is [Reviewable](https://reviewable.io/reviews/m-lab/k8s-support/719)

### DIFF
--- a/config/prometheus/alerts.yml
+++ b/config/prometheus/alerts.yml
@@ -277,10 +277,12 @@ groups:
       dashboard: https://grafana.mlab-oti.measurementlab.net/d/_fugwnWZk/ops-tactical-and-sre-overview?orgId=1&viewPanel=22
 
   # Configuration of the "fq" qdisc on the primary network interface failed, or
-  # the metric is missing for a machine.
+  # the metric is missing for a machine. Don't fire the alert if the node's
+  # status in k8s is "NotReady", or if the node is in maintenance mode.
   - alert: PlatformCluster_ConfigureQdiscFailedOrMissing
     expr: |
       kube_node_info{node=~"mlab[1-4].*"} unless on(node) node_configure_qdisc_success == 1
+        unless on (node) kube_node_status_condition{condition="Ready", status!="true"} == 1
         unless on(node) gmx_machine_maintenance == 1
     for: 1h
     labels:


stephen-soltesz (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/m-lab/k8s-support/719#-:-N9CtXfhB5O5yIEKDyIP:bnfp4nl)* status: 0 of 1 approvals obtained

___
*[`config/prometheus/alerts.yml` line 284 at r1](https://reviewable.io/reviews/m-lab/k8s-support/719#-N9CqzRA29czRjh1FxeF:-N9CqzRA29czRjh1FxeG:b-od9te9) ([raw file](https://github.com/m-lab/k8s-support/blob/22bbeec6d4851a5250115800435124227b0bef0e/config/prometheus/alerts.yml#L284)):*
> ```YAML
>   - alert: PlatformCluster_ConfigureQdiscFailedOrMissing
>     expr: |
>       kube_node_info{node=~"mlab[1-4].*"} unless on(node) node_configure_qdisc_success == 1
> ```

I take it that `kube_node_info` is present for all machines regardless of whether the node is online?

Part of this might be simpler if `node_configure_qdisc_success` was the left hand side metric. It would be present or not only if the node was online. That may make the 'missing' case harder or at least not simpler..

Thinking about these things is not easy.

nkinkade (commented):
> *[Reviewable](https://reviewable.io/reviews/m-lab/k8s-support/719)* status: 0 of 1 approvals obtained

___
*[`config/prometheus/alerts.yml` line 284 at r1](https://reviewable.io/reviews/m-lab/k8s-support/719#-N9CqzRA29czRjh1FxeF:-N9HyREi9QxZZU7gyaXJ:b-orj4yl) ([raw file](https://github.com/m-lab/k8s-support/blob/22bbeec6d4851a5250115800435124227b0bef0e/config/prometheus/alerts.yml#L284)):*

I take it that `kube_node_info` is present for all machines regardless of whether the node is online?

Part of this might be simpler if `node_configure_qdisc_success` was the left hand side metric. It would be present or not only if the node was online. That may make the 'missing' case harder or at least not simpler..

Thinking about these things is not easy.
</blockquote></details>

Yes, `kube_node_info` will exit for a node as long as there is a node object for the machine in the API server.

We talked on the Open Channel VC, and decided that putting the "qdisc" metric on the left would solve the case of a machine being down, but would make discovery of a missing metric more difficult, and would probably end up making the query even more complicated. In this particular case, for this metric, the "absent" case is more important than usual because it is entirely possible that the systemd service that writes out the metric file could fail and not write the file, causing the metric file to not exist _and_ for the fq qdisc to remain unconfigured, configured improperly.



_Code quote:_
```YAML
      kube_node_info{node=~"mlab[1-4].*"} unless on(node) node_configure_qdisc_success == 1
        unless on (node) kube_node_status_condition{condition="Ready", status!="true"} == 1
        unless on(node) gmx_machine_maintenance == 1
```

PR #725, opened 2022-08-24
Adds an alert for when ndt-server can't enable BBR

nkinkade:
This change is [Reviewable](https://reviewable.io/reviews/m-lab/k8s-support/725)

### DIFF
--- a/config/prometheus/alerts.yml
+++ b/config/prometheus/alerts.yml
@@ -297,9 +297,9 @@ groups:
   # module is not loaded.
   - alert: PlatformCluster_BBRmoduleMissing
     expr: |
-      ndt7_measurer_bbr_enabled_total{error="no such file or directory"} > 0
+      increase(ndt7_measurer_bbr_enabled_total{error="no such file or directory"}[5m]) > 0
         unless on(node) gmx_machine_maintenance == 1
-    for: 10m
+    for: 15m
     labels:
       repo: ops-tracker
       severity: ticket
--- a/config/prometheus/alerts.yml
+++ b/config/prometheus/alerts.yml
@@ -293,6 +293,21 @@ groups:
       summary: Configuration of the fq qdisc failed, or the metric is missing.
       description: https://github.com/m-lab/ops-tracker/wiki/Alerts-&-Troubleshooting#platformcluster_configureqdiscfailedormissing
 
+  # The ndt-server is unable to enable BBR for ndt7 tests because the tcp_bbr
+  # module is not loaded.
+  - alert: PlatformCluster_BBRmoduleMissing
+    expr: |
+      ndt7_measurer_bbr_enabled_total{error="no such file or directory"} > 0
+        unless on(node) gmx_machine_maintenance == 1
+    for: 10m
+    labels:
+      repo: ops-tracker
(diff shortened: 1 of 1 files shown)


stephen-soltesz (commented):
> *[Reviewable](https://reviewable.io/reviews/m-lab/k8s-support/725)* status: 0 of 1 approvals obtained (waiting on @nkinkade)

___
*[`config/prometheus/alerts.yml` line 300 at r1](https://reviewable.io/reviews/m-lab/k8s-support/725#-NAGV2pd0oosSHXzGms-:-NAGV2pd0oosSHXzGms0:b-akyey2) ([raw file](https://github.com/m-lab/k8s-support/blob/817044f68bcebde51d088d370cad98e581ab67ea/config/prometheus/alerts.yml#L300)):*
> ```YAML
>   - alert: PlatformCluster_BBRmoduleMissing
>     expr: |
>       ndt7_measurer_bbr_enabled_total{error="no such file or directory"} > 0
> ```

I think a range vector is still preferable here, in case we start a vm and later load the module, we'd like for this condition to age out without other manual intervention (e.g. delete pod to restart it). What do you think?

nkinkade (commented):
> *[Reviewable](https://reviewable.io/reviews/m-lab/k8s-support/725)* status: 0 of 1 approvals obtained (waiting on @stephen-soltesz)

___
*[`config/prometheus/alerts.yml` line 300 at r1](https://reviewable.io/reviews/m-lab/k8s-support/725#-NAGV2pd0oosSHXzGms-:-NAGa_Cj3mTp3ITZ8l9P:bi3x1cx) ([raw file](https://github.com/m-lab/k8s-support/blob/817044f68bcebde51d088d370cad98e581ab67ea/config/prometheus/alerts.yml#L300)):*

I think a range vector is still preferable here, in case we start a vm and later load the module, we'd like for this condition to age out without other manual intervention (e.g. delete pod to restart it). What do you think?
</blockquote></details>

I guess that makes sense. I don't think it should really ever happen that a ndt pod should start up before a kernel module is properly loaded. But I've wrapped the metric in `increase([5m[)` and set the alert threshold for 15m. Does it seem a bit more lenient, yet still effectively, now?

stephen-soltesz (approved):
Thank you.  :lgtm: 

> *[Reviewable](https://reviewable.io/reviews/m-lab/k8s-support/725#-:-NAGzJzI-huVXP-4gPqc:b-mq0xdi)* status: 0 of 1 approvals obtained

___
*[`config/prometheus/alerts.yml` line 300 at r1](https://reviewable.io/reviews/m-lab/k8s-support/725#-NAGV2pd0oosSHXzGms-:-NAGzDHd7lYZgP6pUgsK:b-837i3v) ([raw file](https://github.com/m-lab/k8s-support/blob/817044f68bcebde51d088d370cad98e581ab67ea/config/prometheus/alerts.yml#L300)):*

I guess that makes sense. I don't think it should really ever happen that a ndt pod should start up before a kernel module is properly loaded. But I've wrapped the metric in `increase([5m[)` and set the alert threshold for 15m. Does it seem a bit more lenient, yet still effectively, now?
</blockquote></details>

Okay.

PR #726, opened 2022-09-02
Add prototype configuration for running NDT in GKE

stephen-soltesz:


stephen-soltesz:
FYI: @nkinkade - this just established that a k8s service could preserve the client IP to the ndt-server. This did not preserve the certificate configuration for ndt. Nor can GKE support BBR today. (but a custom image could).

PR #727, opened 2022-09-08
Adds "lame-duck" tolerations to several core DaemonSets

nkinkade:
We use a node taint with key "lame-duck" as a way to sometimes take a node out of production. The issue is that this taint will also cause certain essential workloads to not be scheduled on the node, if they weren't already running there. This PR adds tolerations for this taint to the `node-exporter`, `cadvisor` and `kured` DaemonSets so that essential metrics don't vanish when a node is in lame-duck mode.

Additionally, I accidentally discovered a week or so ago that GCP has apparently enabled IPv6 on GCE VMs in pretty much all GCP regions. To accommodate for this, this PR also includes changes to unconditionally create all virtual platform nodes with a dual stack configuration.

Finally, I discovered that the "n1" machine-type may not be available in all regions (e.g., us-east5), but "n2" should be, so this PR sets the default VM type to `n2-highcpu-4`.

- - -
This change is [Reviewable](https://reviewable.io/reviews/m-lab/k8s-support/727)

### DIFF
--- a/k8s/daemonsets/core/cadvisor.jsonnet
+++ b/k8s/daemonsets/core/cadvisor.jsonnet
@@ -69,6 +69,13 @@
             ],
           },
         ],
+        tolerations: [
+          {
+            effect: 'NoSchedule',
+            key: 'lame-duck',
+            operator: 'Exists',
+          },
+        ],
         volumes: [
           {
             hostPath: {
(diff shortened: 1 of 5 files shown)


SaiedKazemi (approved):
:lgtm:

> Reviewed 5 of 5 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/m-lab/k8s-support/727#-:-NBTmxzL7c-wTTIxuOrj:bnfp4nl)* status: 0 of 1 approvals obtained

PR #729, opened 2022-09-20
Make fluent-bit fetch pod metadata from kubelet instead of api server

nkinkade:
fluent-bit has a feature of its kubernetes filter which will cause it to [fetch metadata about a pod from the local kubelet](https://docs.fluentbit.io/manual/pipeline/filters/kubernetes#optional-feature-using-kubelet-to-get-metadata) instead of from an API server.  This is supposed to alleviate excessive load on the API server(s) in large clusters. Our clusters are not necessarily that large, but the nodes are globally distributed and generally very far away from the k8s API cluster in GCP. The less each fluent-bit pod needs to contact the API server over very long paths the better.

We were previously unable to use this feature because it [required `hostNetwork=true`](https://github.com/fluent/fluent-bit/issues/5143) so fluent-bit could communicate with the Kubelet on 127.0.0.1. However, recently a change was merged which allows users to configure the kubelet host address, which we can easily get from the downward API.

This PR implements this feature, and modifies the fluent-bit DaemonSet to inject the node's public IP into the fluent-bit configuration for `Kubelet_Host`.

Quick note on the `tls.verify Off` configuration. The kubelet's _serving_ certificate is self signed (this is distinct from the cert the kubelet uses to communicate with the API server). We are guaranteed that fluent-bit will be communicating with the public IP of the primary network interface on the same node where fluent-bit is running, so this should be fairly safe.

This PR also updates fluent-bit to v1.9.8.

- - -
This change is [Reviewable](https://reviewable.io/reviews/m-lab/k8s-support/729)


cristinaleonr (approved):
:lgtm: 

> *[Reviewable](https://reviewable.io/reviews/m-lab/k8s-support/729#-:-NCVMmI30vo6x5tX005d:bg7igkv)* status: 0 of 1 approvals obtained

PR #732, opened 2022-10-04
Configurations necessary for upgrade to k8s v1.22.15

nkinkade:
Most of the changes in this PR are related to upgrading k8s to v1.22.15.

The other two unrelated changes are:

* Sets fluent-bit's kubernetes filter setting `Buffer_Size` to 0, which means the buffer can grow as large as needed. We had a situation where a bunch of evicted pods were causing the kubelet's response to its `/pods` endpoint to return more than 1MB of data, causing the filter to fail.
* Creates directory /etc/kubernetes/manifests when bootstrapping a Prometheus VM to avoid log-polluting messages about that directory not existing.

- - -
This change is [Reviewable](https://reviewable.io/reviews/m-lab/k8s-support/732)

### DIFF
--- a/config/fluentbit/values.yaml.template
+++ b/config/fluentbit/values.yaml.template
@@ -138,7 +138,7 @@ initContainers:
     [FILTER]
         Name kubernetes
         Match kube.*
-        Buffer_Size 2MB
+        Buffer_Size 0
         Use_Kubelet true
         Kubelet_Host ${KUBELET_HOST}
         tls.verify Off
--- a/config/fluentbit/values.yaml.template
+++ b/config/fluentbit/values.yaml.template
@@ -138,7 +138,7 @@ initContainers:
     [FILTER]
         Name kubernetes
         Match kube.*
-        Buffer_Size 1MB
+        Buffer_Size 2MB
         Use_Kubelet true
         Kubelet_Host ${KUBELET_HOST}
         tls.verify Off
--- a/manage-cluster/bootstrap_prometheus.sh
+++ b/manage-cluster/bootstrap_prometheus.sh
@@ -232,4 +232,8 @@ EOF2
       # TODO: replace with native k8s persistent volumes, if possible.
       chown nobody:nogroup /mnt/local/prometheus/
   fi
+
+  # The kubelet will complain about this directory not existing every few
+  # seconds, terribly polluting the logs.
+  mkdir -p /etc/kubernetes/manifests
 EOF
--- a/manage-cluster/k8s_deploy.conf
+++ b/manage-cluster/k8s_deploy.conf
@@ -28,18 +28,18 @@ PROM_BASE_NAME="prometheus-${GCE_BASE_NAME}"
 # we learn more.
 GCE_API_SCOPES="cloud-platform"
 
-K8S_VERSION="v1.21.14"
+K8S_VERSION="v1.22.15"
 K8S_CNI_VERSION="v1.1.1"
-K8S_CRICTL_VERSION="v1.21.0"
+K8S_CRICTL_VERSION="v1.22.1"
 # FLANNEL is for the flannel DaemonSet, whereas FLANNELCNI is for the CNI
 # plugin located at /opt/cni/bin (used by the kubelet).
-K8S_FLANNEL_VERSION="v0.19.0"
+K8S_FLANNEL_VERSION="v0.19.2"
 K8S_FLANNELCNI_VERSION=v1.1.0
 K8S_TOOLING_VERSION="v0.14.0"
-ETCDCTL_VERSION="v3.4.14"
-K8S_HELM_VERSION="v3.7.1"
+ETCDCTL_VERSION="v3.5.5"
+K8S_HELM_VERSION="v3.10.0"
 K8S_FLUENTBIT_VERSION="1.9.8-debug"
-K8S_CERTMANAGER_VERSION="v1.6.1"
+K8S_CERTMANAGER_VERSION="v1.9.0"
 K8S_CERTMANAGER_DNS01_SA="cert-manager-dns01-solver"
 K8S_CERTMANAGER_SA_KEY="cert-manager-credentials.json"
 K8S_CA_FILES="ca.crt ca.key sa.key sa.pub front-proxy-ca.crt front-proxy-ca.key etcd/ca.crt etcd/ca.key"
--- a/manage-cluster/kubeadm-config.yml.template
+++ b/manage-cluster/kubeadm-config.yml.template
@@ -1,15 +1,16 @@
-apiVersion: kubeadm.k8s.io/v1beta2
+apiVersion: kubeadm.k8s.io/v1beta3
 kind: InitConfiguration
 nodeRegistration:
   name: "{{MASTER_NAME}}"
   kubeletExtraArgs:
     container-runtime: "remote"
     container-runtime-endpoint: "unix:///run/containerd/containerd.sock"
+  criSocket: "unix:///run/containerd/containerd.sock"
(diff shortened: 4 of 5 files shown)


cristinaleonr (commented):
> *[Reviewable](https://reviewable.io/reviews/m-lab/k8s-support/732)* status: 0 of 1 approvals obtained (waiting on @nkinkade)

___
*[`manage-cluster/bootstrap_prometheus.sh` line 238 at r1](https://reviewable.io/reviews/m-lab/k8s-support/732#-NDZNMRM9sHUE8YUHVAR:-NDZNMRN9xofLivP4ZYM:b-wq453o) ([raw file](https://github.com/m-lab/k8s-support/blob/908d131412e1145dc92433e52a0f167ca84e2bd8/manage-cluster/bootstrap_prometheus.sh#L238)):*
> ```Shell
>   # The kubelet will complain about this directory not existing every few
>   # seconds, terribly polluting the logs.
>   mkdir -p /etc/kubernetes/manifests
> ```

What does the kubelet want to use this directory for?

nkinkade (commented):
> *[Reviewable](https://reviewable.io/reviews/m-lab/k8s-support/732)* status: 0 of 1 approvals obtained (waiting on @cristinaleonr)

___
*[`manage-cluster/bootstrap_prometheus.sh` line 238 at r1](https://reviewable.io/reviews/m-lab/k8s-support/732#-NDZNMRM9sHUE8YUHVAR:-NDZOKoL7csC4zSB99d2:b-29nfi1) ([raw file](https://github.com/m-lab/k8s-support/blob/908d131412e1145dc92433e52a0f167ca84e2bd8/manage-cluster/bootstrap_prometheus.sh#L238)):*
<details><summary><i>Previously, cristinaleonr (Cristina Leon) wrote…</i></summary><blockquote>

What does the kubelet want to use this directory for?
</blockquote></details>

[Static pod](https://kubernetes.io/docs/tasks/configure-pod-container/static-pod/) manifests. I believe we only use them on the control-plane nodes. The kubelet config on each node defines a `staticPodPath`  configuration. The kubelet checks that path for pod manifests that it should be managing locally, and will log a warning when it doesn't exist. 

Your question had me wondering what sets `staticPodPath` in the kubelet configuration, and for the life of me I was unable to figure out what sets that. When nodes join the cluster using `kubeadm join`, kubeadm downloads the kubelet configuration from a configmap in the cluster. All kubelet-configuration configmaps going back to v1.17 have that config set to /etc/kubernetes/manifests. One _possible_ way to get around creating that directory to avoid log warnings would be to remove that setting from our kubelet configuration.

cristinaleonr (commented):
> *[Reviewable](https://reviewable.io/reviews/m-lab/k8s-support/732)* status: 0 of 1 approvals obtained (waiting on @nkinkade)

___
*[`manage-cluster/bootstrap_prometheus.sh` line 238 at r1](https://reviewable.io/reviews/m-lab/k8s-support/732#-NDZNMRM9sHUE8YUHVAR:-ND_TCVD73NZiFOdNmZ9:by6vhq8) ([raw file](https://github.com/m-lab/k8s-support/blob/908d131412e1145dc92433e52a0f167ca84e2bd8/manage-cluster/bootstrap_prometheus.sh#L238)):*

[Static pod](https://kubernetes.io/docs/tasks/configure-pod-container/static-pod/) manifests. I believe we only use them on the control-plane nodes. The kubelet config on each node defines a `staticPodPath`  configuration. The kubelet checks that path for pod manifests that it should be managing locally, and will log a warning when it doesn't exist. 

Your question had me wondering what sets `staticPodPath` in the kubelet configuration, and for the life of me I was unable to figure out what sets that. When nodes join the cluster using `kubeadm join`, kubeadm downloads the kubelet configuration from a configmap in the cluster. All kubelet-configuration configmaps going back to v1.17 have that config set to /etc/kubernetes/manifests. One _possible_ way to get around creating that directory to avoid log warnings would be to remove that setting from our kubelet configuration.
</blockquote></details>

If it's necessary for the management of static pods, then we should leave it. Though I wonder if the kubelet ever had any problems managing the static pods when the directory didn't exist.

If it's possible to change the configuration such that it won't need/look for the path, then that would make the most sense IMO.

nkinkade (commented):
> *[Reviewable](https://reviewable.io/reviews/m-lab/k8s-support/732)* status: 0 of 1 approvals obtained (waiting on @cristinaleonr)

___
*[`manage-cluster/bootstrap_prometheus.sh` line 238 at r1](https://reviewable.io/reviews/m-lab/k8s-support/732#-NDZNMRM9sHUE8YUHVAR:-NDd4ZQC5kxP6cckdq9Q:b8mffjb) ([raw file](https://github.com/m-lab/k8s-support/blob/908d131412e1145dc92433e52a0f167ca84e2bd8/manage-cluster/bootstrap_prometheus.sh#L238)):*
<details><summary><i>Previously, cristinaleonr (Cristina Leon) wrote…</i></summary><blockquote>

If it's necessary for the management of static pods, then we should leave it. Though I wonder if the kubelet ever had any problems managing the static pods when the directory didn't exist.

If it's possible to change the configuration such that it won't need/look for the path, then that would make the most sense IMO.
</blockquote></details>

`staticPodPath` or the (deprecated) `--pod-manifest-path` flag are the two ways to configure the kubelet to manage static pods. Lacking one of those configurations the kubelet won't manage any static pods. If one of those are configured, but the path doesn't exist, then the kubelet will function just fine, but will complain ever few seconds that it does not exist, polluting logs. This change is just to get the kubelet to stop polluting logs, not to make it manage static pods any better.

I'm hesitant to remove the `staticPodPath` configuration directive from the kubelet-config configmap, since, as far as I know, kubeadm uses that configuration for control plane nodes as well as cluster nodes, and that configuration is imperative for control plane nodes. There  are currently only 3 types of nodes in the cluster:

* control plane nodes
* physical nodes
* virtual nodes

Creation of that directory is already handled  for most all of those types of node, with the one exception of the Prometheus VM, which this PR addresses.

cristinaleonr (approved):
:lgtm: 

> *[Reviewable](https://reviewable.io/reviews/m-lab/k8s-support/732#-:-NDe2_xu65tOhk0YFv2V:bg7igkv)* status: 0 of 1 approvals obtained (waiting on @nkinkade)

___
*[`manage-cluster/bootstrap_prometheus.sh` line 238 at r1](https://reviewable.io/reviews/m-lab/k8s-support/732#-NDZNMRM9sHUE8YUHVAR:-NDe2PE_3x5LwBD2eIfK:b71z3jw) ([raw file](https://github.com/m-lab/k8s-support/blob/908d131412e1145dc92433e52a0f167ca84e2bd8/manage-cluster/bootstrap_prometheus.sh#L238)):*

`staticPodPath` or the (deprecated) `--pod-manifest-path` flag are the two ways to configure the kubelet to manage static pods. Lacking one of those configurations the kubelet won't manage any static pods. If one of those are configured, but the path doesn't exist, then the kubelet will function just fine, but will complain ever few seconds that it does not exist, polluting logs. This change is just to get the kubelet to stop polluting logs, not to make it manage static pods any better.

I'm hesitant to remove the `staticPodPath` configuration directive from the kubelet-config configmap, since, as far as I know, kubeadm uses that configuration for control plane nodes as well as cluster nodes, and that configuration is imperative for control plane nodes. There  are currently only 3 types of nodes in the cluster:

* control plane nodes
* physical nodes
* virtual nodes

Creation of that directory is already handled  for most all of those types of node, with the one exception of the Prometheus VM, which this PR addresses.
</blockquote></details>

Okay, if the configuration shouldn't be removed, then let's leave it as is and add the path.

PR #736, opened 2022-10-24
Configures ndt-virtual to allow all ndt5 without tokens

nkinkade:
Now that virtual nodes are becoming more of a mainstay on the platform, virtual nodes should handle ndt5 traffic in the same way as physical nodes in order to avoid silent errors for legacy clients that don't know how to use access tokens, or otherwise don't use the Locate Service. In other words, this PR should cause ndt-server on virtual nodes to treat all clients in the same way that ndt-server on physical nodes does.


SaiedKazemi (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/m-lab/k8s-support/736#-:-NFAvAiVD60WbGf0Raqj:bnfp4nl)* status: 0 of 1 approvals obtained

PR #741, opened 2022-11-15
Adds memory limit of 1500M to pcap on virtual nodes

nkinkade:
This PR is a stab at a temporary workaround for this issue: https://github.com/m-lab/ops-tracker/issues/1682

I'm not very happy with this PR... adding new function parameters, etc. It WAI, but seems ugly. Additionally, I'm not fully convinced that this will 100% resolve the issue. It may keep the VM from crashing, however. Another side effect could _possibly_ be that the kernel will repeatedly oom-kill the packet-headers container, which will cause the kubelet to flag the ndt-virtual pod as `NotReady`, which may cause an alert to fire.

- - -
This change is [Reviewable](https://reviewable.io/reviews/m-lab/k8s-support/741)

### DIFF
--- a/k8s/daemonsets/experiments/ndt-virtual.jsonnet
+++ b/k8s/daemonsets/experiments/ndt-virtual.jsonnet
@@ -6,7 +6,7 @@ local services = [
   'ndt/ndt5=ws://:3001/ndt_protocol,wss://:3010/ndt_protocol',
 ];
 
-exp.ExperimentNoIndex(expName, 'pusher-' + std.extVar('PROJECT_ID'), 'none', datatypes, true) + {
+exp.ExperimentNoIndex(expName, 'pusher-' + std.extVar('PROJECT_ID'), 'none', datatypes, true, 'virtual') + {
   metadata+: {
     name: expName + '-virtual',
   },
--- a/k8s/daemonsets/templates.jsonnet
+++ b/k8s/daemonsets/templates.jsonnet
@@ -209,7 +209,7 @@ local Traceroute(expName, tcpPort, hostNetwork) = [
   else []
 ;
 
-local Pcap(expName, tcpPort, hostNetwork) = [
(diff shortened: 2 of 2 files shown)


nkinkade (commented):
> *[Reviewable](https://reviewable.io/reviews/m-lab/k8s-support/741)* status: 0 of 1 approvals obtained (waiting on @stephen-soltesz)

___
*[`k8s/daemonsets/templates.jsonnet` line 246 at r1](https://reviewable.io/reviews/m-lab/k8s-support/741#-NGxdaho8-jNz4CrAeQU:-NGxvAET2no5AaD0dli0:bjhvpin) ([raw file](https://github.com/m-lab/k8s-support/blob/94d88dafa28d42b42ea5237c451e8e83169ec21a/k8s/daemonsets/templates.jsonnet#L246)):*

Is this the value you determined was typical? What metric specifically?

My instinct from earlier this week (though without looking at the actual data) was 2GB.
</blockquote></details>

Yeah, we had conversationally talked about 50% of the memory, which would be roughly 2GB. However, once I looked at some data, it _appeared_ that in some cases pcap spiking above even 1G was sometimes causing a brief outage of metrics, but not crashing the VM. I just went back to look at things again and see that for all the VMs that actually did crash, they were all at some point above 2G of memory usage. I also notice that there have been some other VMs where memory usage went above 2GB, causing an outage of metrics, but not apparently crashing the VM. Maybe it has to do with longevity of the SYN flood and not the intensity. I've upped the limit to 2G. If we continue to have crashed VMs, then we can lower that value.

stephen-soltesz (approved):
:lgtm:

> Reviewed 1 of 2 files at r1.
> *[Reviewable](https://reviewable.io/reviews/m-lab/k8s-support/741#-:-NGy0NBH4zOpEKUUKCfE:bnfp4nl)* status: 0 of 1 approvals obtained

PR #749, opened 2022-12-13
Updates fluent-bit to v2.0.6

nkinkade:
Also, take advantage of some new "threaded" feature, whereby input plugins can run in a seprate thread and not back up the main thread. Also, reduces storage.total_limit_size to 250M. Our usage of container logs in Stackdriver is more of a nice-to-have than mission-critical-and-we-cannot-lose-even-1byte-of-log-data. We don't need to buffer 1G of logs if there are network or backpressure issues, and 250M should be adequate.

Also, maybe there are some fixes to the issues we have had where fluentbit encounters network issues, then just stops doing anything.

- - -
This change is [Reviewable](https://reviewable.io/reviews/m-lab/k8s-support/749)


cristinaleonr (approved):
:lgtm: 

> *[Reviewable](https://reviewable.io/reviews/m-lab/k8s-support/749#-:-NJD0n4l8CHzJ46koyTo:bg7igkv)* status: 0 of 1 approvals obtained

PR #757, opened 2023-01-17
Explicitly sets JSON credential file path

nkinkade:
Previously we were setting the env variable GOOGLE_APPLICATION_CREDENTIALS. The documentation for this sink says that in the absence of this configuration it will check that env variable. And indeed it appears to work that way, but over the weekend around 8 Vector pods started throwing errors about invalid tokens, when we don't even use tokens to authenticate Vector. This made me wonder if there could be some bug in Vector whereby it was not checking that env variable, or checking it too late, and falling back to some other authentication method that fails. This commit explicitly sets the path to the JSON key for authentication, which is probably a better configuration even if it doesn't turn out to solve any problems.

- - -
This change is [Reviewable](https://reviewable.io/reviews/m-lab/k8s-support/757)

### DIFF
--- a/helm/vector/values.yaml.template
+++ b/helm/vector/values.yaml.template
@@ -5,6 +5,7 @@ customConfig:
   sinks:
     stackdriver:
       type: gcp_stackdriver_logs
+      credentials_path: "/etc/vector/keys/vector.json"
       inputs:
       - kernel_log
       - kubernetes_logs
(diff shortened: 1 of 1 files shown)


cristinaleonr (approved):
It might be helpful to check if anyone has raised any issues about these errors in Vector's repo.

:lgtm: 

> *[Reviewable](https://reviewable.io/reviews/m-lab/k8s-support/757#-:-NM0ok7NAz_FXjtpj7kX:b-z39tth)* status: 0 of 1 approvals obtained

nkinkade (commented):
I looked earlier, and there are [a few issues](https://github.com/vectordotdev/vector/issues?q=is%3Aissue+unauthorized+is%3Aopen) about getting 401 responses from GCP, but they mostly seem to have to do with people likely deploying Vector inside of GCP and who are actually relying on tokens to authenticate. 

> *[Reviewable](https://reviewable.io/reviews/m-lab/k8s-support/757#-:-NM0pBls8bFLSLHbkTZz:bjmmaq4)* status: 0 of 1 approvals obtained

PR #761, opened 2023-01-26
Adds scripts k8s scripts to initialize the control plane

nkinkade:
Currently, the control plane is bootstrapped via [a collection](https://github.com/m-lab/k8s-support/blob/main/manage-cluster/bootstrap_platform_cluster.sh) of [shell scripts](https://github.com/m-lab/k8s-support/blob/main/manage-cluster/bootstraplib.sh), which are run manually by an operator. Workloads are deployed by two scripts [create_k8s_configs.sh](https://github.com/m-lab/k8s-support/blob/main/manage-cluster/create_k8s_configs.sh) and [apply_k8s_configs.sh](https://github.com/m-lab/k8s-support/blob/main/manage-cluster/apply_k8s_configs.sh). After the cluster is bootstrapped, changes to the workloads are deployed by Cloud Builds, running these same "create" and "apply" scripts.

The current scripts are not suitable for running in an automated fashion directly on control plane nodes when a cluster is initialized. This PR contains essentially those same two scripts, but modified slightly to make them suitable for being run by a script on the control plane nodes when the cluster is bootstrapped automatically.

These changes are meant to be a stop-gap until we can find a way to have both cluster initialization and cluster updates use the same scripts. For now, the scripts should be functionally equivalent, resulting in the same workloads, but using different variables sourced from different locations, and some of the steps are reordered.

- - -
This change is [Reviewable](https://reviewable.io/reviews/m-lab/k8s-support/761)

### DIFF
--- a/helm/cert-manager/helm-values-overrides.yaml
+++ b/helm/cert-manager/helm-values-overrides.yaml
@@ -6,8 +6,12 @@ cainjector:
 ingressShim:
   defaultIssuerName: letsencrypt
   defaultIssuerKind: ClusterIssuer
+installCRDs: true
 nodeSelector:
   "node-role.kubernetes.io/control-plane": ""
+startupapicheck:
+  tolerations:
+  - operator: Exists
 strategy:
   type: Recreate
 tolerations:
@@ -17,3 +21,4 @@ webhook:
     "node-role.kubernetes.io/control-plane": ""
   tolerations:
   - operator: Exists
+
--- a/helm/cert-manager/helm-values-overrides.yaml
+++ b/helm/cert-manager/helm-values-overrides.yaml
@@ -6,7 +6,6 @@ cainjector:
 ingressShim:
   defaultIssuerName: letsencrypt
   defaultIssuerKind: ClusterIssuer
-installCRDs: true
 nodeSelector:
   "node-role.kubernetes.io/control-plane": ""
 startupapicheck:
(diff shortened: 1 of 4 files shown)


stephen-soltesz (approved):
:lgtm:

Just clarifying - when you say above that these scripts are "suitable for being run by a script on the control plane nodes" - do you mean running them on the same node as will become a Kubernetes cluster api server?

If so, my first thought is whether there could be a temporary node/environment created to run these scripts to target the Kubernetes cluster so as not to cross the streams of cluster bootstrap and cluster configuration, maybe like how we use CloudBuild today but perhaps natively in GCE so it could be managed by Terraform? This is 1/2 an idea so if it doesn't make sense feel free to leave it.

> *[Reviewable](https://reviewable.io/reviews/m-lab/k8s-support/761#-:-NMoPfIq5Id-Pd8CxeG9:bwyaso9)* status: 0 of 1 approvals obtained

nkinkade:
@stephen-soltesz: After reflecting on your comments in this PR and our discussion in the Open VC about the epoxy-images/Packer PR, and further thinking this through myself, I have decided to close this PR. Populating the cluster with workloads does indeed feel like a separate concern from creating the control plane. For now, I'll let the scripts initialize the control plane, but leave it empty of M-Lab workloads, to be populated by some other method. One thought is that builds of terraform-support, could perhaps use `cbctl` to trigger a k8s-support build, which would deploy workloads to the empty cluster, for example. There may be other strategies.

PR #770, opened 2023-02-16
Updates k8s to v1.23.16 + updates related components and services

nkinkade:
* A bare `--dry-run` has been deprecated for a long time now, and is removed in v1.23.
* `node-cidr-mask-size` is still a supported flag the kube-controller, but now that dual-stack clusters are GA, there are two new protocol-specific flags... that same flag with `-ipv4` and `-ipv6` suffixes. Just to be explicit, I'm choosing to use the new flag, rather than an implicit one.
* Control plane machine names (now that they are controlled by Terraform in sandbox) no longer begin with "master-", but instead "api-".
* Mounting a GCS bucket using gcsfuse is huge overkill for the purpose of simply uploading a single file a bucket. Instead, just us gsutil to upload the file.

- - -
This change is [Reviewable](https://reviewable.io/reviews/m-lab/k8s-support/770)

### DIFF
--- a/manage-cluster/add_k8s_api_node.sh
+++ b/manage-cluster/add_k8s_api_node.sh
@@ -111,7 +111,7 @@ gcloud compute ssh "${BOOTSTRAP_MASTER}" "${GCP_ARGS[@]}" --zone "${BOOTSTRAP_MA
   kubectl create configmap kubeadm-config -n kube-system \
       --from-file ClusterConfiguration \
       --from-file ClusterStatus \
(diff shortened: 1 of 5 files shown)


robertodauria (approved):
:lgtm:

> Reviewed 5 of 5 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/m-lab/k8s-support/770#-:-NOozKkeAqY87xWbuCwo:bnfp4nl)* status: 0 of 1 approvals obtained

PR #777, opened 2023-03-07
Install Kured via helm instead of static manifests

nkinkade:
This PR install Kured via Helm instead of static manifests. The problem with static manifests is that they are just that, static. We may use a sample/recommended manifest, and then a few years may go by. During that time, the recommended configuration may have changed, but we are still stuck with the old, static config. With Kured being managed by Helm, whenever we upgrade Kured  to a new version, Helm should automatically drag in all the new related resources (Roles, RoleBindings, etc.).

Additionally, I changed how Vector's version is handled to match how I handled Kured's. Namely, the version of the service is now set via the Helm `--set` flag rather than injecting the right version into a values-overrides template.

- - -
This change is [Reviewable](https://reviewable.io/reviews/m-lab/k8s-support/777)


robertodauria (approved):
:lgtm:

> Reviewed 9 of 9 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/m-lab/k8s-support/777#-:-NPyJ-oE2gYAW7j1A052:bnfp4nl)* status: 0 of 1 approvals obtained

PR #785, opened 2023-03-22
Runs most containers as non-root

nkinkade:
This PR causes most containers to run as user `nobody:nogroup` (which on Debian-based systems seems to be 65534:65534). 

There are some containers that still need to run as root:

* packet-headers
* ndt-virtual (ndt-server container)
* reponsiveness (responsiveness-server container)

... in these cases, the containers still run as root, but with only the capabilities needed to function.

Since most experiment data volumes are of type `hostPath`, we need to change the ownership of the data directories to nobody:nogroup. `hostPath` directories will get the effective user of the kubelet (i.e., root) by default. To accomplish this, I have included an initContainer that runs as root and changes the ownership of the directory.

There is an additional initContainer that handles setting the ownership of `/var/spool/datatypes` for autoloading experiments. 

The Wehe container image version is updated to use a newer version which supports a standard M-Lab filesystem layout, allowing Wehe to run as non-root.

This PR also reenables required envelope access tokens for the Wehe experiment.

- - -
This change is [Reviewable](https://reviewable.io/reviews/m-lab/k8s-support/785)


stephen-soltesz:
Part of:
* https://github.com/m-lab/ops-tracker/issues/1650

PR #787, opened 2023-03-27
Deploy MSAK daemonset in staging

robertodauria:
This PR adds a conditional to system.jsonnet to deploy MSAK's `latest` image from Docker Hub in staging, too. This will give people working on MSAK clients a larger platform to experiment with than just sandbox servers, ever if there is no tagged release of msak-server yet.

- - -
This change is [Reviewable](https://reviewable.io/reviews/m-lab/k8s-support/787)

### DIFF
--- a/k8s/daemonsets/experiments/msak.jsonnet
+++ b/k8s/daemonsets/experiments/msak.jsonnet
@@ -37,7 +37,7 @@ exp.Experiment(expName, 1, 'pusher-' + std.extVar('PROJECT_ID'), "none", datatyp
                 },
               },
             ],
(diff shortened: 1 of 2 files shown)


stephen-soltesz:
@robertodauria I don't object to deploying to staging, but I'm concerned about relying on `latest` in staging. As you continue to develop msak the build backing that version will update and staging nodes could end up running a mixture of versions (though all labeled 'latest'). Is it possible to make the image tag conditional on project?

robertodauria:
@stephen-soltesz I tagged `v0.0-alpha1` and updated this PR so the image deployed is stable. Thank you.

PR #802, opened 2023-05-22
Do not drop privileges in ndt-virtual's ndt containers

robertodauria:
For the past few days, the ndt-server instances running on virtual nodes were not able to write to the output folder due to permission issues. This temporary workaround was put in place by @stephen-soltesz through manual editing of the ndt-virtual DS in production.

This PR aligns the ndt-virtual DS with what's currently running in production so we can deploy k8s-support while looking for a better long-term solution.

- - -
This change is [Reviewable](https://reviewable.io/reviews/m-lab/k8s-support/802)

PR #804, opened 2023-05-23
Fix ndt-virtual's permission issue

robertodauria:
The ndt/ndt5 and ndt/ndt7 folders are created by Kubernetes as soon as the pod is created, with owner root/root and permissions 0755.

The set-data-dir-perms initContainer then sets the owner of /var/spool/ndt to nobody/nogroup (recursively) and permissions to `2775` (775 + setgid bit) but does not do so for subfolders, leaving the ndt/ndt5 and ndt/ndt7 with their original 0755 permissions which only allow the `nobody` user to write there. 

In the ndt-virtual DS, ndt-server runs as capability-less root (so, uid=0 but no root privileges), thus it cannot write to the output folder.

This PR fixes that by:
1. Applying the 2775 permissions to subfolders, so ndt/ndt5 and ndt/ndt7 become group-writable
2. Restoring the `drop: all` that was introduced by #802 as a workaround

### DIFF
--- a/k8s/daemonsets/experiments/ndt-virtual.jsonnet
+++ b/k8s/daemonsets/experiments/ndt-virtual.jsonnet
@@ -99,7 +99,6 @@ exp.ExperimentNoIndex(expName, 'pusher-' + std.extVar('PROJECT_ID'), 'none', dat
                 ],
               },
               runAsUser: 0,
-              runAsGroup: 65534,
             },
             volumeMounts: [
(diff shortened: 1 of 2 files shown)


nkinkade (approved):
:lgtm:

stephen-soltesz:
Part of:
* https://github.com/m-lab/ops-tracker/issues/1650

PR #823, opened 2023-06-29
Explicit separation of data and datatypes volumes for all services

stephen-soltesz:
Previously, the jostler datatypes volume definition referenced `/var/spool/datatypes` within the host filesystem. This was a mistake, since all service data should be stored under `/cache/*`. As well, all services used a shared data volume, so all schema files were placed in the same directory visible to any other service that also mounted the datatypes volume. This violates service isolation, because (for example) one service could corrupt or overwrite the schema of a second service, which should be impossible.

This change updates the templates.jsonnet configuration to place the host datatypes directory under `/cache/datatypes/<name>` and this also partitions the location for every service. From within the pod namespace, the location will remain unchanged as `/var/spool/datatypes`.

- - -
This change is [Reviewable](https://reviewable.io/reviews/m-lab/k8s-support/823)


stephen-soltesz:
FYI: @cristinaleonr

PR #849, opened 2023-08-29
Exclude all k8s experiment container logs from GCP 

nkinkade:
https://github.com/m-lab/k8s-support/issues/847

While working on the above fixit week issue, I realized that pretty much all logs from experiment containers are not helpful for almost every normal use case, except debugging, which we rarely have to do. Even in the cases where we do need to debug, and logs are useful, the logs exist in the cluster and those should generally be enough. 

This PR add the label `vector.dev/exclude: true` to all experiment pods, which means that Vector will not upload logs to stackdriver for any experiment pod until this label is removed or set to "false". In some special event that we want container logs for a pod in GCP, then we can either do a PR/release cycle, or just manually edit the DaemonSet in place changing "true" to "false" for that label.

- - -
This change is [Reviewable](https://reviewable.io/reviews/m-lab/k8s-support/849)

PR #850, opened 2023-09-01
Enables exporting of internal Vector metrics 

nkinkade:
We recently [stopped pushing experiment container logs to GCP](https://github.com/m-lab/k8s-support/pull/849), and as part of that [we removed the alert `ContainerLogsMissingIngInStackdriver`](https://github.com/m-lab/prometheus-support/pull/1007). However, we have consistently found that Vector containers can get into a broken state with no visible symptom other than it stops pushing container logs successfully. Vector still does push kernel logs and _some_ low  volume container logs (e.g., cAdvisor, node-exporter, kube-state-metrics, etc). We still need some sort of alert for when Vector is broken, yet the container doesn't crash (which itself would cause an alert over time). 

This PR enables the `internal_metrics` data source and pushes it the `prometheus_exporter` sink on port 9090. Vector exports [a ridiculous number of metrics](https://vector.dev/docs/reference/configuration/sources/internal_metrics/#output-metrics), and we are likely only concerned with a few. However, until we can discover a broken container and find out which metric(s) are relevant, we collect all metrics exported by Vector. At some point we can either use Vector itself to filter some metric or we can filter on the Prometheus side.

- - -
This change is [Reviewable](https://reviewable.io/reviews/m-lab/k8s-support/850)

### DIFF
--- a/helm/vector-values-overrides.yaml.template
+++ b/helm/vector-values-overrides.yaml.template
@@ -7,6 +7,7 @@ customConfig:
       type: prometheus_exporter
       inputs:
       - vector_metrics
+	  address: 0.0.0.0:9090
     stackdriver:
       type: gcp_stackdriver_logs
       credentials_path: "/etc/vector/keys/vector.json"
@@ -37,8 +38,8 @@ customConfig:
 env:
 - name: VECTOR_LOG
   value: info
-extraContainerPorts:
-- 9598
+extraContainersPorts:
+- 9090
 extraVolumes:
 - name: credentials
   secret:
--- a/helm/vector-values-overrides.yaml.template
+++ b/helm/vector-values-overrides.yaml.template
@@ -25,8 +25,8 @@ customConfig:
       type: journald
     kubernetes_logs:
       type: kubernetes_logs
-	vector_metrics:
-	  type: vector_metrics
+    vector_metrics:
+      type: vector_metrics
   transforms:
     kernel_log_filter:
       type: filter
--- a/helm/vector-values-overrides.yaml.template
+++ b/helm/vector-values-overrides.yaml.template
@@ -26,7 +26,7 @@ customConfig:
     kubernetes_logs:
       type: kubernetes_logs
     vector_metrics:
-      type: vector_metrics
+      type: internal_metrics
   transforms:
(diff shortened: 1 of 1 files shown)


stephen-soltesz (approved):
I'd advocate for the vector-filtered configuration in the future if that's what we need rather than separating the vector config and the prometheus configs. That would be a little more self-contained I think.

:lgtm:

> *[Reviewable](https://reviewable.io/reviews/m-lab/k8s-support/850#-:-NdGqEfD6-VyuH-7-XMQ:b-aeqty8)* status: 0 of 1 approvals obtained (waiting on @nkinkade)

___
*[`helm/vector-values-overrides.yaml.template` line 10 at r1](https://reviewable.io/reviews/m-lab/k8s-support/850#-NdGq9pp4Ep9B_ZWEuOy:-NdGq9pp4Ep9B_ZWEuOz:b-qv996y) ([raw file](https://github.com/m-lab/k8s-support/blob/2f77b70a545a4e02b9bda6c31566ca7b6cf33426/helm/vector-values-overrides.yaml.template#L10)):*
> ```
>       inputs:
>       - vector_metrics
>       address: 0.0.0.0:9090
> ```

Just confirming vector is listening on a private pod network only?

nkinkade (commented):
I'm sure there is a way to filter metrics in vector through some transform or other. We can cross that bridge when we get to it.

> *[Reviewable](https://reviewable.io/reviews/m-lab/k8s-support/850#-:-NdHX2TeA7pXaxa-y8Tn:b-wjsbun)* status: 0 of 1 approvals obtained (waiting on @stephen-soltesz)

___
*[`helm/vector-values-overrides.yaml.template` line 10 at r1](https://reviewable.io/reviews/m-lab/k8s-support/850#-NdGq9pp4Ep9B_ZWEuOy:-NdHSVuP5zNMdXHAuYjp:b-renb2g) ([raw file](https://github.com/m-lab/k8s-support/blob/2f77b70a545a4e02b9bda6c31566ca7b6cf33426/helm/vector-values-overrides.yaml.template#L10)):*

Just confirming vector is listening on a private pod network only?
</blockquote></details>

Vector does not have a public address, and only gets the normal private pod address.

PR #862, opened 2023-12-13
Removes conditional for packet-headers memory limit flags

nkinkade:
Previously, we restricted applying the flags `-maxidleram` and `-maxheap` to only VMs. This change applies those flags to all machine types, both VMs and physical. The expected outcome is that this shouldn't modify behavior of packet-headers on physical machines most all of the time, but may prevent packet-headers from eating up too much memory and invoking the oom-killer due to SYN floods or heavy traffic spikes. In either case packet-headers data is corrupted or non-existent, but with the flags it may not take the machine down or make it unresponsive.

- - -
This change is [Reviewable](https://reviewable.io/reviews/m-lab/k8s-support/862)


stephen-soltesz (approved):
> *[Reviewable](https://reviewable.io/reviews/m-lab/k8s-support/862)* status: 0 of 1 approvals obtained

PR #874, opened 2024-04-09
Upgrade K8S_VERSION to v1.27.12 + related components

robertodauria:
This PR changes the k8s version to 1.27.12. I've also upgraded other related components, like crictl and CNI, etc. as per the [upgrade instructions](https://docs.google.com/document/d/1SUl-D3eDGFG5XsWSkq0eVcUahuMG7HIFucezmvbI6kM/edit#heading=h.ev2qge8jgv0x). The sandbox control plane has been upgraded already.

The [Cloud Build](https://console.cloud.google.com/cloud-build/builds;region=global/8f495805-20e4-4550-830a-645e61f9d639;step=2?project=mlab-sandbox) completed with no errors

Marking this PR as a draft to make sure it's not merged until the api-platform-cluster nodes have been upgraded in staging, but please consider it reviewable.

- - -
This change is [Reviewable](https://reviewable.io/reviews/m-lab/k8s-support/874)

PR #875, opened 2024-04-09
Fix Vector and flannel container image tags

nkinkade:
In a [previous PR](https://github.com/m-lab/k8s-support/pull/874) the `-debian` suffix was accidentally removed from the version string for Vector. This PR adds that back.

Additionally, before, flannel released container image tags with an `-amd64` suffix, but it appears that they stopped doing this not long ago, so this PR removes that suffix.

- - -
This change is [Reviewable](https://reviewable.io/reviews/m-lab/k8s-support/875)

### DIFF
--- a/k8s/daemonsets/core/flannel-physical.jsonnet
+++ b/k8s/daemonsets/core/flannel-physical.jsonnet
@@ -106,7 +106,7 @@
             command: [
               'cp',
             ],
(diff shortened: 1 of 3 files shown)


robertodauria (approved):
:lgtm:

> Reviewed 3 of 3 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/m-lab/k8s-support/875#-:-Nv4LEAKEQBsnI50PiG1:bnfp4nl)* status: 0 of 1 approvals obtained

PR #902, opened 2024-09-11
Condenses flannel to a single DaemonSet

nkinkade:
This is a companion PR to https://github.com/m-lab/epoxy-images/pull/268

As noted in the above PR, currently, we have a weird configuration in the k8s platform cluster with two separate flannel DaemonSets, one for physical machines and one for virtual machines. The whole point of this is writing out different CNI config files depending on the machine type. With the CNI configs baked into the boot images, flannel no longer needs to do this task in an initContainer. And because we build separate images for physical and virtual, it becomes trivial to just drop the right CNI config into the rootfs of the image.

This should serve to simplify our cluster configurations, with relatively small drawback that a small part of cluster configuration is now handled by epoxy-images rather than there in k8s-support. That said, we haven't really changed those CNI config files in years, so they are are more or less static for the foreseeable future.


stephen-soltesz (approved):
> *[Reviewable](https://reviewable.io/reviews/m-lab/k8s-support/902)* status: 0 of 1 approvals obtained (waiting on @nkinkade)

___
*[`config/flannel.jsonnet` line 17 at r1](https://reviewable.io/reviews/m-lab/k8s-support/902#-O6X1N2r3mBJRxi2QC-I:-O6X1N2r3mBJRxi2QC-J:b-8obkxn) ([raw file](https://github.com/m-lab/k8s-support/blob/7478157cebb546b0e4c62c0bdf069add43ce09be/config/flannel.jsonnet#L17)):*
> ```Jsonnet
> local virtualCNIConf = {
>   name: 'cbr0',
>   cniVersion: '0.2.0',
> ```

Just noting that the cni version from https://reviewable.io/reviews/m-lab/epoxy-images/268 was 0.3.1 also.

nkinkade (commented):
> *[Reviewable](https://reviewable.io/reviews/m-lab/k8s-support/902)* status: 0 of 1 approvals obtained (waiting on @stephen-soltesz)

___
*[`config/flannel.jsonnet` line 17 at r1](https://reviewable.io/reviews/m-lab/k8s-support/902#-O6X1N2r3mBJRxi2QC-I:-O6XeS6aAUlAcMpQK83b:b-ta2d1g) ([raw file](https://github.com/m-lab/k8s-support/blob/7478157cebb546b0e4c62c0bdf069add43ce09be/config/flannel.jsonnet#L17)):*

Just noting that the cni version from https://reviewable.io/reviews/m-lab/epoxy-images/268 was 0.3.1 also.
</blockquote></details>

Good catch. I may have done that without even thinking. There was this comment in this file for `virtualCNIConf`:

```
// cniVersion:0.2.0 is the CNI version that index2ip knows about. It needn't be
// this older version on virtual nodes, but we keep it there to be the same as
// the physical nodes (see physicalCNIConf below).
```

I don't recall doing it consciously, but I may have just opted to make it 0.3.1, since it didn't really matter anyway.

PR #915, opened 2024-11-05
Change -flowtimeout to 5s

robertodauria:
This changes the value of `-flowtimeout` from the default (30s) to 5s. As a result, a flow will be considered stopped, and its corresponding goroutine will be terminated and garbage-collected after 5 seconds of no new packets being observed.

See https://github.com/m-lab/dev-tracker/issues/800

- - -
This change is [Reviewable](https://reviewable.io/reviews/m-lab/k8s-support/915)

### DIFF
--- a/k8s/daemonsets/templates.jsonnet
+++ b/k8s/daemonsets/templates.jsonnet
@@ -361,6 +361,7 @@ local Pcap(expName, tcpPort, hostNetwork, siteType, anonMode) = [
       '-anonymize.ip=' + anonMode,
       '-maxidleram=1500MB',
       '-maxheap=2GB',
+      '-flowtimeout=5s',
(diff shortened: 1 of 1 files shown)


stephen-soltesz (approved):
Is there any open issue with a summary of the rationale motivating this change? Including a link to that issue in the description would provide a breadcrumb to anyone who wanted to understand why this is needed and why it could be wise to preserve in the future, or at least to have more context about what went into the decision in case someone wants to change it in the future with less context.

:lgtm:

> *[Reviewable](https://reviewable.io/reviews/m-lab/k8s-support/915#-:-OAwswHa62Xf_oXy7LCE:ba733tb)* status: 0 of 1 approvals obtained

PR #922, opened 2025-02-24
Create new system for implementing nftables rules in experiment pods

nkinkade:
Part of this is removing all the MultiNetworkPolicy manifests and configurations, which decided to abandon because:

* it took down the production k8s API cluster when we deployed it to production
* the upstream code seems abandoned and there is no likelihood of support and future development

This new system is comprised of a simple set of nftables rules for each experiment. Those rules are then added to a new ConfigMap named `nftables-rules`. Each experiment pod will mount the particular data key for itself from the ConfigMap to /etc/nftables.d/mlab.conf and apply those rules as part of an initContainer in the pod.

The ports that get opened should be the same ones from the previous MultiNetworkPolicy configurations. Additionally, ports 9990-9999 are opened to allow Prometheus to get metrics from all the various sidecars. Also, all ICMP traffic is explicitly allowed.

- - -
This change is [Reviewable](https://reviewable.io/reviews/m-lab/k8s-support/922)


robertodauria (approved):
:lgtm:

> Reviewed 18 of 18 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/m-lab/k8s-support/922#-:-OK2E6lHDyGPSQl9X-LZ:bnfp4nl)* status: 0 of 1 approvals obtained

nkinkade:
This was closed by accident, when I mistakenly deleted the sandbox-kinkade branch when doing some other work in the repo. I'll get it back.

nkinkade:
Reopen this PR.

PR #944, opened 2026-02-10
Removes the packet-test experiment

nkinkade:
The experiment produced the data that was needed for the paper that was written a year or more ago, and we no longer need to be running it on the platform.

- - -
This change is [Reviewable](https://reviewable.io/reviews/m-lab/k8s-support/944)


robertodauria (approved):
Thanks!

